PR #4, opened 2020-09-03
shim-sev: Cargo.lock update

haraldh:


haraldh:
Do we want `Cargo.lock` in git?

connorkuehl:
From what I've read on the internet, it seems the rule of thumb is to check in the `Cargo.lock` for binaries, but not for libraries. [source](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html).

IIRC, the shim exposes a library component to share type definitions with the loader which is pretty library-ish, so I don't know what the right answer is here.

haraldh:
> IIRC, the shim exposes a library component to share type definitions with the loader which is pretty library-ish, so I don't know what the right answer is here.

From your source:
> If you’re building a non-end product, such as a rust library that other rust packages will depend on, put Cargo.lock in your .gitignore.

Which is the case here.. we don't build a library for others to consume.

npmccallum:
Yes, we definitely want `Cargo.lock`. @haraldh Can you file an issue to add a test to check that the `Cargo.lock` files are up-to-date?

npmccallum (approved):

connorkuehl (approved):

haraldh:
> Yes, we definitely want `Cargo.lock`. @haraldh Can you file an issue to add a test to check that the `Cargo.lock` files are up-to-date?

@npmccallum done
https://github.com/enarx/enarx-keepldr/issues/7

PR #5, opened 2020-09-03
Add getting started documentation

npmccallum:


lkatalin (approved):

src/main.rs
```diff
@@ -1,6 +1,50 @@
...
-//! Loads keeps from their backends.
+//! This crate provides the `enarx-keepldr` executable which loads `static-pie`
+//! binaries into an Enarx Keep - that is a hardware isolated environment using
+//! technologies such as Intel SGX or AMD SEV.
+//!
+//! # Install Dependencies
+//!
+//! ## Fedora
+//!
+//!     $ sudo dnf install git curl gcc pkg-config openssl-devel musl-gcc
```
haraldh, line 11:
````suggestion
//! ```console
//! $ sudo dnf install git curl gcc pkg-config openssl-devel musl-gcc
//! ```
````

haraldh: and the rest likewise ... just a suggestion

haraldh (approved):

PR #11, opened 2020-09-08
Establish basis for modularizing VM construction

connorkuehl:
This allows backends to further specialize VM construction to meet their
own needs without having to duplicate any VM construction code in any of
the VM-based backends.

A backend can participate in the construction of a VM by implementing
the vm_base::builder::Hook trait and overriding some (or all) of the
methods in that trait to perform the work it needs to do.

The vm_base::builder::Hook trait can be easily extended to add additional
steps if needed.

No functional changes in terms of the VM's personality; this is just a
large refactoring.

Unused abstractions, such as the Allocator type from cpu.rs, have been
removed. Furthermore, some things that were needlessly coupled, such as
the VMSetup struct were also removed (since this type is only used
during 1-time setup).


src/backend/vm_base/builder.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub fn new(cfg: Config, shim: Component, code: Component, hook: Option<Box<dyn Hook>>) -> Self {
+        Self {
+            hook: hook.unwrap_or_else(|| Box::new(NoHook)),
+            shim,
+            code,
+            config: cfg,
+        }
+    }
+
+    pub fn build(mut self) -> Result<Vm> {
```
npmccallum, line 66: This function is too large. Please split it into smaller functions.

src/backend/vm_base/builder.rs
```diff
@@ -0,0 +1,198 @@
...
+
+pub struct NoHook;
+impl Hook for NoHook {}
+
+pub struct Builder {
+    hook: Box<dyn Hook>,
+    shim: Component,
+    code: Component,
+    config: Config,
+}
```
npmccallum, line 54: Do you need to own the hook? Or is it sufficient just to borrow it? Also, why dynamic dispatch here? Is this impossible?

```suggestion
pub struct Builder<T: Hook> {
    hook: T,
    shim: Component,
    code: Component,
    config: Config,
}
```

npmccallum (changes requested):

src/backend/mod.rs
```diff
@@ -3,6 +3,7 @@
 pub mod kvm;
 pub mod sev;
 pub mod sgx;
+mod vm_base;
```
npmccallum, line 6: I'm not sure I'm a fan of this. Can we just export these types from the `kvm` module?

src/backend/kvm/vm/builder.rs
```diff
@@ -1,284 +1,243 @@
...
+    pub fn load_segments(start: VirtAddr, segs: &[Segment]) {
+        use std::slice::from_raw_parts_mut;
 
-        Ok(vm)
+        for seg in segs {
+            let dst = VirtAddr::new(seg.dst as u64 + start.as_u64());
+            let dst = unsafe { from_raw_parts_mut(dst.as_mut_ptr::<Page>(), seg.src.len()) };
+            dst.copy_from_slice(&seg.src[..]);
+        }
     }
```
npmccallum, line 242: Merge into `load_component()`.

npmccallum (changes requested):

src/backend/vm_base/builder.rs
```diff
@@ -0,0 +1,243 @@
...
+            *setup.as_mut_ptr::<PageTable>() = PageTable::new();
+            pml4t = &mut *setup.as_mut_ptr::<PageTable>();
+            setup += size_of::<PageTable>();
+
+            *setup.as_mut_ptr::<PageTable>() = PageTable::new();
+            pml3t_ident = &mut *setup.as_mut_ptr::<PageTable>();
+            setup += size_of::<PageTable>();
+        }
+
+        let start = VirtAddr::new(map.addr() as u64);
```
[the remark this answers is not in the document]
connorkuehl: Do you mind if I file this as a debt issue? Most of our stuff is `usize`, however, `x86_64` deals only in `u64`

haraldh (approved):

src/backend/kvm/mod.rs
```diff
@@ -1,20 +1,22 @@
 // SPDX-License-Identifier: Apache-2.0
 
+mod builder;
 mod shim;
 mod vm;
 
+pub const SHIM: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/bin/shim-sev"));
```
npmccallum, line 7: I might be missing it. But I don't think `SHIM` is actually used anywhere else. Does it need to be public?

haraldh: I would guess, not yet and will be used in a follow-up PR

connorkuehl: Correct ^ -- I can change it for these patches though

npmccallum (approved):

PR #12, opened 2020-09-09
shim-sev: handle SEV encryption

haraldh:


internal/shim-sev/src/start.S
```diff
@@ -25,6 +25,52 @@
...
 _start:
+.intel_syntax noprefix
+CheckSevFeature:
+    // Check if we have a valid (0x8000_001F) CPUID leaf
+    mov       eax, 0x80000000
+    cpuid
+
+    // This check should fail on Intel or Non SEV AMD CPUs. In future if
+    // Intel CPUs supports this CPUID leaf then we are guranteed to have exact
+    // same bit definition.
```
npmccallum, line 36: I'm skeptical about this claim.

npmccallum (changes requested):
Two larger concerns:

1. The amount of asm has grown significantly. We now have a megafunction written in assembly that is very hard to follow. Can this be broken up?

2. I thought we agreed on making the C-bit a constant with an assert if it changes.

internal/shim-sev/src/start.S
haraldh: Ooops, shamelessly copied.

Author: Brijesh Singh <brijesh.singh@amd.com>

https://github.com/AMDESE/ovmf/commit/e60af8a1ebb15bfcbf2ecc4afb6cf35084c847aa

haraldh:
> Two larger concerns:
> 
>     1. The amount of asm has grown significantly. We now have a megafunction written in assembly that is very hard to follow. Can this be broken up?
> 
>     2. I thought we agreed on making the C-bit a constant with an assert if it changes.

1. IMHO, nothing before  `call    _dyn_reloc` can be made into rust safely
2. wasn't that big of a deal and a hardcoded version wouldn't change anything

haraldh:
>     1. The amount of asm has grown significantly. We now have a megafunction written in assembly that is very hard to follow. Can this be broken up?

the sgx `start.S` has an equal amount of lines

internal/shim-sev/src/start.S
haraldh: want me to remove the comments?

npmccallum (approved):

connorkuehl (approved):

PR #13, opened 2020-09-09
Support for daemon, enarx-wasmldr

MikeCamel:
This is a big merge from enarx/enarx.  Requires docs.


npmccallum:
@MikeCamel This code doesn't belong here. It belongs in `enarx-keepmgr`.

MikeCamel:
> @MikeCamel This code doesn't belong here. It belongs in `enarx-keepmgr`.

Which bits of this code?  enarx-keepmgr holds code for the central keep manager process, which manages and coordinates Keep loader instanes.  Anything which is per instance lives here: that's how I segmented the code (as we agreed, I believe).  If we want to have a different split, we need to discuss how that works. 

I can see moving all of the daemon pieces out into enarx-keepmgr makes sense if the idea is for keepldr to hold only implementations of Keeps.  Is that what you had in mind?

npmccallum:
> > @MikeCamel This code doesn't belong here. It belongs in `enarx-keepmgr`.
> 
> Which bits of this code? enarx-keepmgr holds code for the central keep manager process, which manages and coordinates Keep loader instanes. Anything which is per instance lives here: that's how I segmented the code (as we agreed, I believe). If we want to have a different split, we need to discuss how that works.

We agreed to a different split. `enarx-keepmgr` holds all the system-wide state. `enarx-wasmldr` holds all the per-instance state. The `enarx-keepldr` has a very minimal job: start and run a static pie binary in a keep. Nothing else.

> I can see moving all of the daemon pieces out into enarx-keepmgr makes sense if the idea is for keepldr to hold only implementations of Keeps. Is that what you had in mind?

Yes. It is supposed to be bare bones. `enarx-keepmgr` and `enarx-wasmldr` need to be integrated to each other. But `enarx-keepldr` should be a (mostly) invisible middle man. This is why I described `enarx-keepmgr` as being able to work by spawning `/usr/bin/enarx-wasmldr` directly **OR** via `enarx-keepldr exec /usr/bin/enarx-wasmldr`. The result should be (mostly) the same.

MikeCamel:
> > > @MikeCamel This code doesn't belong here. It belongs in `enarx-keepmgr`.
> > Which bits of this code? enarx-keepmgr holds code for the central keep manager process, which manages and coordinates Keep loader instances. Anything which is per instance lives here: that's how I segmented the code (as we agreed, I believe). If we want to have a different split, we need to discuss how that works.
> 
> We agreed to a different split. `enarx-keepmgr` holds all the system-wide state. `enarx-wasmldr` holds all the per-instance state. The `enarx-keepldr` has a very minimal job: start and run a static pie binary in a keep. Nothing else.

**NOTE** - please read the below, but don't get irate until you've read the "Important bit" (which will hopefully mean you don't need to get irate).  I kept what I've written in between because I think it raises some useful points in the meantime.

I think we may have agreed to different things.  :-)  Part of the reason for the design as I've done it to this point is to allow state to be held by keeploader so that it can be pre-warmed at an early stage, and manage comms in a (guaranteed) multi-threaded manner with the keepmanager or other possible entities.  This allows us the broadest set of options for different attestation, billing and other protocols in the future, and why there's a state machine in here.  It's also why I wasn't planning on losing the originating process (though I think I'm probably OK with this once that wasmtime piece is loaded and running).  

This, btw, is why I wanted to go through the protocol design and work out what we needed, rather than designing on the fly: doing this allows a data model for exactly what data is owned by which components, and when.  It seems that we've come up with conflicting ideas of what sits where, and how best to manage future options, and we need to come to an agreement or we'll go through this again: there are multiple moving parts which need coordination, not stand-alone components, and building distributed systems needs more design work.  In particular, I still think we need to write out exactly what the early SEV protocol version is, as it's an important corner case.

> > I can see moving all of the daemon pieces out into enarx-keepmgr makes sense if the idea is for keepldr to hold only implementations of Keeps. Is that what you had in mind?
 
> Yes. It is supposed to be bare bones. `enarx-keepmgr` and `enarx-wasmldr` need to be integrated to each other. But `enarx-keepldr` should be a (mostly) invisible middle man. This is why I described `enarx-keepmgr` as being able to work by spawning `/usr/bin/enarx-wasmldr` directly **OR** via `enarx-keepldr exec /usr/bin/enarx-wasmldr`. The result should be (mostly) the same.

Design principle: don't replicate functionality.  The keeploader (or associated daemon - see below) should be the thing that knows how to create and pre-provision Keeps (including "nil Keeps" or "runtimes which can take workloads").  The keepmgr interacts with the client to gather enough information to launch a Keep, which job it hands off to the keeploader.  The client then talks to the Keep.

Design principle: maintain state close to where it's needed, and allow it to be owned by the component which needs it and can change it most directly.  Keep-specific state should be managed individually, not centrally (this also allows for recovery in the case of a failed keepmanager.)

I'm not saying that we necessarily need the daemon model, but we need to be sure we _don't_ need it before we design it out, and I'm not sure that the (possibly) single-threaded nature of some Keeps means that the wasmldr is where this sits.

**Important bit** - on re-reading what I've written and what you've written, I'm unclear how far apart we actually are.  I think my main concerns are around a) multi-threading in wasmldr & b) possible protocol futures.  It _may_ be that all you're advocating is moving the daemon part out of enarx-keepldr into enarx-keepmgr, in which case that's pretty easy.  If so, then please acknowledge!

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "enarx-keepldr"
 version = "0.1.0"
-authors = ["Nathaniel McCallum <npmccallum@redhat.com>"]
+authors = ["Nathaniel McCallum <npmccallum@redhat.com>", "Mike Bursell <mike@p2ptrust.org>"]
```
haraldh, line 4: that must be important.. tsk, tsk

npmccallum:
> **NOTE** - please read the below, but don't get irate until you've read the "Important bit" (which will hopefully mean you don't need to get irate). I kept what I've written in between because I think it raises some useful points in the meantime.

Nobody's irate. :)

> **Important bit** - on re-reading what I've written and what you've written, I'm unclear how far apart we actually are. I think my main concerns are around a) multi-threading in wasmldr & b) possible protocol futures. It _may_ be that all you're advocating is moving the daemon part out of enarx-keepldr into enarx-keepmgr, in which case that's pretty easy. If so, then please acknowledge!

I don't think we're far apart. 

The design of `keepldr` is to be an adapter. It makes keeps look like normal Linux processes. It has to have a minimal amount of specialized knowledge in order to handle the `pre-attestation` workflow. But that should be it. Other than that, its job is to make things look normal and then get out of the way.

We agree that state should be held close. This is why per-instance state is in `wasmldr` (the closest it can be!) and per-host state is in `keepmgr`.

I'm not worried about the lack of threading in `wasmldr`. In fact, I think avoiding threading there altogether is desirable. Using `tokio` (as we're doing) helps us here **significantly**. With `tokio` we can handle multiple clients at the same time without threads, which is really all we need.

PR #20, opened 2020-09-15
Add madvise() to shim-sev

npmccallum:


connorkuehl (approved):

internal/shim-sev/src/syscall.rs
```diff
@@ -663,6 +664,15 @@ impl Handler {
             }
         }
     }
+
+    pub fn madvise(&self) -> Result<usize, libc::c_int> {
+        eprintln!(
+            "SC> madvise(0x{:x}, 0x{:x}, 0x{:x}) = 0",
```
haraldh, line 670: If you want prefix with `0x` use `#x`

haraldh (approved):

PR #29, opened 2020-10-01
cargo update

haraldh:
SSIA


haraldh:
Somehow this could be automated...

```
$ cargo update
$ ( cd internal/shim-sev; cargo update )
$ ( cd internal/shim-sgx; cargo update )

PR #32, opened 2020-10-01
Enable dependabot version updates for Cargo

### LINKED ISSUE #30: Enable Dependabot
While we have Dependabot enabled, it is not configured, so it's not doing anything for us. We should add a configuration file to allow it to do its thing.

mbestavros:
As discussed in today's standup. It's currently set to run daily.

Resolves #30.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+version: 2
+updates:
+    - package-ecosystem: "cargo"
+        directory: "/"
+        schedule:
+            interval: "daily"
+    - package-ecosystem: "cargo"
+        directory: "internal/rcrt1"
+        schedule:
+            interval: "daily"
+    - package-ecosystem: "cargo"
+        directory: "internal/sallyport"
+        schedule:
+            interval: "daily"
+    - package-ecosystem: "cargo"
+        directory: "internal/shim-sev"
+        schedule:
+            interval: "daily"
+    - package-ecosystem: "cargo"
+        directory: "internal/shim-sgx"
+        schedule:
+            interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    # Check the npm registry for updates every day (weekdays)
```
connorkuehl, line 5: ~~`# Check the npm registry for updates every day (weekdays)`~~

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
```
connorkuehl, line 4: We need this to also run for the crates under `internal/`

connorkuehl (changes requested):

.github/dependabot.yml
npmccallum: To be clear, this is the only repo that has this problem.

.github/dependabot.yml
mbestavros: Whoops. Fixed, thanks!

.github/dependabot.yml
mbestavros: I've just added a second `cargo` entry with a second directory path. I think that's how you're supposed to do it.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+    - package-ecosystem: "cargo"
+        directory: "/"
+        schedule:
+            interval: "daily"
+    - package-ecosystem: "cargo"
+        directory: "internal/"
+        schedule:
+        interval: "daily"
```
npmccallum, line 10: This is still not correct.

```suggestion
    - package-ecosystem: "cargo"
        directory: "internal/rcrt1"
        schedule:
            interval: "daily"
    - package-ecosystem: "cargo"
        directory: "internal/sallyport"
        schedule:
            interval: "daily"
    - package-ecosystem: "cargo"
        directory: "internal/shim-sev"
        schedule:
            interval: "daily"
    - package-ecosystem: "cargo"
        directory: "internal/shim-sgx"
        schedule:
            interval: "daily"
```

npmccallum (changes requested):

npmccallum (approved):

connorkuehl (approved):

PR #50, opened 2020-10-05
Remove python `cc` linker stub

haraldh:
Add the `compiler_builtins` crate for the missing memory functions.

Factor out the heap sgx module to its own crate, to run the tests.

Add the linker flags to use our own rcrt1.

Fixes: https://github.com/enarx/enarx/issues/899


internal/no-std-stub/src/lib.rs
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
```
connorkuehl, line 1: Instead of implementing {bcmp, memcmp, memcpy, memmove, memset} could we depend on [compiler_builtins](https://crates.io/crates/compiler_builtins) w/ the [`mem`](https://docs.rs/compiler_builtins/0.1.36/compiler_builtins/mem/index.html) feature as a dependency? Since it provides them, the linker should be able to do the right thing.

haraldh: done

connorkuehl (approved):

npmccallum (approved):

internal/no-std-stub/src/lib.rs
haraldh: @connorkuehl @npmccallum: I just saw that `compiler_builtins` needs `nightly` with quite a lot of `features`. When we move to stable we have to revert using it.

Issue #14, opened 2020-09-11
Document memlock rlimit requirement for SEV-capable hosts


connorkuehl:
The default maximum number of pinnable pages is 16 (at least it was on our Fedora box). A system administrator must configure this setting to be larger on a host where enarx-keepldr will be used with the SEV backend.

We configured ours with an entry under `/etc/security/limits.d/`:

`/etc/security/limits.d/sev.conf`:

```console
* - memlock 8388608
```

```console
$ man limits.conf
```

npmccallum:
@dagrh This has implications for qemu too. Any suggestions on a default policy that works for everyone?

dagrh:
@npmccallum I think libvirt normally sets the rlimit for us when the entry is set in the XML for the VM:

https://libvirt.org/kbase/launch_security_sev.html#memory

PR #52, opened 2020-10-05
Incorporate memlock rlimit check into SEV-readiness

connorkuehl:
Closes: #14 

Looks like this:

```console
$ cargo run info
[..]
 ✔ MEMLOCK rlimit allows for: ~1638 keeps (soft limit = 8589934592 bytes, hard limit = 8589934592 bytes)
```

Or on a machine that can't meet this requirement:

```console
$ cargo run info
[..]
 ✗ MEMLOCK rlimit allows for: 0 keeps (soft limit = 65536 bytes, hard limit = 65536 bytes)

The MEMLOCK rlimit must be large enough to accommodate the Enarx shim, wasmldr, and the memory pressure requirements of the target workloads across all deployed SEV keeps.
```

### DIFF
--- a/src/backend/sev/mod.rs
+++ b/src/backend/sev/mod.rs
@@ -12,7 +12,7 @@ use anyhow::Result;
 
 use std::arch::x86_64::__cpuid_count;
 use std::fs::OpenOptions;
-use std::mem::transmute;
+use std::mem::{transmute, MaybeUninit};
 use std::path::Path;
 use std::str::from_utf8;
 use std::sync::{Arc, RwLock};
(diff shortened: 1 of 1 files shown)


haraldh:
Does that mean, it will automatically silently fallback to plain `kvm`, if the memlock rlimit is too low?

I am thinking of the testsuite.

connorkuehl:
> Does that mean, it will automatically silently fallback to plain `kvm`, if the memlock rlimit is too low?
> 
> I am thinking of the testsuite.

For CI, no, because the CI sets the `ENARX_BACKEND` variable which will fail if the backend checks fail.

For a regular developer workflow running `cargo test` on their own machine, yes, this will silently switch to `kvm` if the developer doesn't set `ENARX_BACKEND`.

connorkuehl:
> The tool should probably be verbose about running a non TEE backend.

Emitting this information on stdout/stderr will spuriously break our integration tests for test runs on a non-TEE backend.

haraldh:
Isn't the backend specified and there is no fallback then?

connorkuehl:
> Isn't the backend specified and there is no fallback then?

Not if you're just running `cargo test` without `ENARX_BACKEND` set. This is pretty common in my workflow.

Silently switching to a supported backend when `ENARX_BACKEND` is not set is how this has always behaved; it isn't introduced by this PR.

haraldh (approved):

lkatalin (approved):
This makes sense!

PR #54, opened 2020-10-06
Set stack size according to rust profile

haraldh:
Set stack size according to rust profile
    
### DIFF
--- a/internal/shim-sev/src/start.S
+++ b/internal/shim-sev/src/start.S
@@ -14,7 +14,11 @@
 
 #define SHIM_OFFSET 0xFFFFFF8000000000
 
-#define SIZE_OF_INITIAL_STACK (128 * 1024)
+#ifdef NDEBUG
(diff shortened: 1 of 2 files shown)


connorkuehl (approved):

internal/shim-sev/src/start.S
```diff
@@ -14,7 +14,11 @@
 
 #define SHIM_OFFSET 0xFFFFFF8000000000
 
-#define SIZE_OF_INITIAL_STACK (128 * 1024)
+#ifdef NDEBUG
+#define SIZE_OF_INITIAL_STACK (48 * 1024)
+#else
+#define SIZE_OF_INITIAL_STACK (150 * 1024)
```
lkatalin, line 20: Just curious why these sizes?

lkatalin (approved):

internal/shim-sev/src/start.S
haraldh: @lkatalin Empirical values... try and error for now

PR #60, opened 2020-10-08
Update x86_64 crate and use new GDT descriptors

haraldh:


lkatalin (commented):
We are okay with this adding new flags from [here](https://github.com/rust-osdev/x86_64/blob/master/src/structures/gdt.rs#L237), right? Looks like most are ignored in 64-bit mode, but there is [`ACCESSED`](https://github.com/rust-osdev/x86_64/blob/master/src/structures/gdt.rs#L189).

haraldh:
accessed is fine for now

lkatalin (approved):

ueno (approved):

PR #61, opened 2020-10-08
build.rs: pickup more rerun-if-changed

haraldh:
Use `walkdir::WalkDir` to traverse the directories.

Add all crates in `./internal/` to rerun-if-changed.


connorkuehl (approved):

lkatalin (approved):
Good idea.

PR #70, opened 2020-10-09
Execute a simple `echo` rust test program.

haraldh:
This builds any `tests/bin/*.rs` found, which can be used for
integration tests.

A simple `echo.rs` test writing every `stdin` to `stdout` is added.

Because this uses `poll`, the `poll` syscall was also implemented for
the `sgx` and `sev` shims.

### DIFF
--- a/internal/shim-sev/src/syscall.rs
+++ b/internal/shim-sev/src/syscall.rs
@@ -134,6 +134,7 @@ pub extern "C" fn syscall_rust(
         libc::SYS_fstat => h.fstat(),
         libc::SYS_fcntl => h.fcntl(),
         libc::SYS_madvise => h.madvise(),
+        libc::SYS_poll => h.poll(),
 
         syscall => {
             //panic!("SC> unsupported syscall: {}", syscall);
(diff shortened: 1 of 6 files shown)


lkatalin (approved):

connorkuehl:
IIRC the reason why we compile C binaries for the integration tests is because it is easy to avoid linking with crt0; the goal is to ensure the scope of the test is not accidentally widened by whatever crt0 may or may not call. From my initial read of this, it doesn't look like the builds for the Rust binaries exclude crt0.

I like the idea of testing real programs, and have no strong opinion as to whether those are included in the integration tests or if we organize them collectively in a different repository. However, this may be a good place to break some ground with another kind of end-to-end testing.

I have been thinking it'd be useful to have some kind of repo/CI setup that is able to clone the relevant repos, build the Enarx stack, and feed in WebAssembly binaries. This setup would be associated with each component that builds up our `end-to-end` run of a Wasm workload. This way, any time we change any component in the runtime stack, they're all tested together.

For example, a pull request made to the `enarx-keepldr` repo would trigger the following:

1. The PR would be checked out
2. The CI would clone the missing components (in this case, `enarx-wasmldr` & the `enarx-test-progs` repo; if the PR was originally made to `enarx-wasmldr`, then the missing components would be `enarx-keepldr` & `enarx-test-progs`, etc.)
3. The CI would build each program in the `enarx-test-progs` designation as a Wasm binary.
3. The CI would use `enarx-keepldr` to exec the `enarx-wasmldr` for each build test program.

This would also the side effect of serving as a demonstration of the gradually more complicated workloads Enarx is able to deploy.

Thoughts?

haraldh:
> Thoughts?

Yeah, I like the `end-to-end` integration test. We should definitely have that.

This test serves more of a quick and lean `end-to-end` rust test and would show regressions with `cargo test` earlier in the devel cycle.

internal/shim-sev/src/syscall.rs
```diff
@@ -696,6 +697,37 @@ impl Handler {
 
         Ok(0)
     }
+
+    pub fn poll(&self) -> Result<usize, libc::c_int> {
+        let nfds = self.b as libc::nfds_t;
+        let timeout = self.c as libc::c_int;
+        let trusted =
+            unsafe { core::slice::from_raw_parts_mut(self.a as *mut libc::pollfd, nfds as _) };
```
npmccallum, line 705: This could be cleaned up by making the registers into `Register` types. But this would effect all the syscalls. Separate PR please.

npmccallum (approved):

PR #71, opened 2020-10-09
shim-sev: unlock hostcall to print PanicInfo

haraldh:
In case of a panic, we force get `HOST_CALL` mutex to print the panic
message.

### DIFF
--- a/internal/shim-sev/src/main.rs
+++ b/internal/shim-sev/src/main.rs
@@ -42,6 +42,7 @@ use primordial::Address;
 use sallyport::Block;
 use spinning::RwLock;
 
+use crate::hostcall::HOST_CALL;
 pub use hostlib::BootInfo;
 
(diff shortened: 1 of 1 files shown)


lkatalin (approved):
PanicInfo is useful and this seems fine to me, though I am not sure of all the implications of forcibly unlocking a mutex. (I did find documentation on `force_unlock` [here](https://docs.rs/lock_api/0.4.1/lock_api/struct.Mutex.html#method.force_unlock)).

haraldh:
> PanicInfo is useful and this seems fine to me, though I am not sure of all the implications of forcibly unlocking a mutex. (I did find documentation on `force_unlock` [here](https://docs.rs/lock_api/0.4.1/lock_api/struct.Mutex.html#method.force_unlock)).

@lkatalin so, yeah... `force_unlock()` is a pretty bad thing do normally, but here we are panicing anyway.

PR #73, opened 2020-10-09
switch away from lazy_static

haraldh:
Use a similar `Lazy` as the experimental `std::lazy::Lazy`.

lazy_static is unmaintained.

### DIFF
--- a/internal/shim-sev/src/frame_allocator.rs
+++ b/internal/shim-sev/src/frame_allocator.rs
@@ -3,6 +3,7 @@
 //! The global FrameAllocator
 use crate::addr::{HostVirtAddr, ShimPhysAddr, ShimPhysUnencryptedAddr, ShimVirtAddr};
 use crate::hostcall::HOST_CALL;
+use crate::lazy::Lazy;
 use crate::{get_cbit_mask, BOOT_INFO};
 
 use nbytes::bytes;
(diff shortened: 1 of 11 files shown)


connorkuehl (approved):

internal/shim-sev/src/frame_allocator.rs
```diff
@@ -23,15 +24,12 @@ use x86_64::{align_down, align_up, PhysAddr, VirtAddr};
...
-    pub static ref FRAME_ALLOCATOR: RwLock<FrameAllocator> = {
-        RwLock::<FrameAllocator>::const_new(
-            spinning::RawRwLock::const_new(),
-            unsafe { FrameAllocator::new() },
-        )
-    };
-}
+/// The global ShimFrameAllocator RwLock
+pub static FRAME_ALLOCATOR: Lazy<RwLock<FrameAllocator>> = Lazy::new(|| {
+    RwLock::<FrameAllocator>::const_new(spinning::RawRwLock::const_new(), unsafe {
```
ambaxter, line 29: Could you use the same formatting breaks as the old code? It was easier to read.

ambaxter (approved):

internal/shim-sev/src/frame_allocator.rs
haraldh: Well, `rustfmt` does all the stuff, which it didn't inside the `lazy_static!` macro call.
I can try to persuade `rustfmt` with local vars to format it more readable.

Issue #59, opened 2020-10-07
Use the `volatile` crate for `Cursor::alloc`


haraldh:
To prevent potential compiler optimizations after a host context switch, reading from and writing to the sallyport block should be declared `volatile`.

https://github.com/rust-osdev/volatile

volatile's `copy_from_slice()` and `copy_into_slice()` currently rely on the currently unstable [`intrinsics::volatile_copy_nonoverlapping_memory`](https://doc.rust-lang.org/core/intrinsics/fn.volatile_copy_nonoverlapping_memory.html) though.

~~With 1.47~~  `VolatileArray` and `VolatileSlice` might be possible according to https://github.com/rust-osdev/volatile/issues/8
Edit: `1.47` only allowed const array generics in `std` ... it's still not accessible for stable.

PR #74, opened 2020-10-12
Improve the sallyport cursor data handling

haraldh:
* shims: Add a compiler memory fence for the host proxy calls

    To prevents the compiler from optimizing read and writes not knowing of
    the side-effects of the syscall, `core::sync::atomic::compiler_fence`
    is added before and after the host call via sallyport.
    
    Therefore this patch also changes the now unnecessary volatile
    writes into normal writes.
    
    Fixes: https://github.com/enarx/enarx-keepldr/issues/59   

* sallyport: Re-enable tests

    Re-enable the test suite by temporarily changing the
    `Block::buf_capacity` to `Page::size()` if tests are enabled,
    otherwise the tests will fail with a buffer overflow.
    
    Also ignore the `syscall` test for `cargo miri test`, because
    inline assembly is not supported for `miri`.

* sallyport: Extend `Cursor` with read/write convenience methods
    
    This patch extends the `Cursor` with read/write convenience
    methods for various use cases, refactoring some code and
    moving unsafe copy operations to a central place.
    
    Because most pointers don't have to be dereferenced,
    this prevents potential UB (undefined behaviour) usage.
    
    Related: https://github.com/enarx/enarx-keepldr/issues/78   

* sallyport: Return MaybeUninit for `Cursor::alloc()`
    
    let `Cursor::alloc()` return `&mut MaybeUninit<T>`
    instead of `&mut [T]` to indicate, that the buffer is uninitialized.

### DIFF
--- a/internal/shim-sev/src/hostcall.rs
+++ b/internal/shim-sev/src/hostcall.rs
@@ -69,7 +69,15 @@ impl<'a> HostCall<'a> {
     #[inline(always)]
     pub unsafe fn hostcall(&mut self) -> sallyport::Result {
         let mut port = Port::<u16>::new(SYSCALL_TRIGGER_PORT);
+
+        // prevent earlier writes from being moved beyond this point
+        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::Release);
+
         port.write(1);
+
+        // prevent earlier reads from being moved before this point
+        core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::Acquire);
+
         self.0.msg.rep.into()
     }
 
--- a/internal/shim-sev/src/hostcall.rs
+++ b/internal/shim-sev/src/hostcall.rs
@@ -8,7 +8,6 @@ use crate::hostlib::{MemInfo, SYSCALL_TRIGGER_PORT, SYS_ENARX_BALLOON_MEMORY, SY
 use crate::lazy::Lazy;
 use crate::SHIM_HOSTCALL_VIRT_ADDR;
 use core::convert::TryFrom;
-use core::mem::MaybeUninit;
 use primordial::{Address, Register};
 use sallyport::{request, Block};
 use spinning::Mutex;
@@ -101,7 +100,7 @@ impl<'a> HostCall<'a> {
     /// The parameters returned can't be trusted.
     pub unsafe fn write(&mut self, fd: usize, bytes: &[u8]) -> sallyport::Result {
         let cursor = self.0.cursor();
-        let (_, buf) = cursor.copy_slice(bytes).or(Err(libc::EMSGSIZE))?;
+        let (_, buf) = cursor.copy_from_slice(bytes).or(Err(libc::EMSGSIZE))?;
 
         let buf_address = Address::from(buf.as_ptr());
         let phys_unencrypted = ShimPhysUnencryptedAddr::try_from(buf_address).unwrap();
@@ -119,19 +118,16 @@ impl<'a> HostCall<'a> {
 
     /// Get host memory info
     pub fn mem_info(&mut self) -> Result<MemInfo, libc::c_int> {
-        let mut mem_info = MaybeUninit::<MemInfo>::uninit();
-
         self.0.msg.req = request!(SYS_ENARX_MEM_INFO);
 
         let _result = unsafe { self.hostcall() }?;
 
         let block = self.as_mut_block();
         let c = block.cursor();
-        let (_, untrusted) = unsafe { c.alloc::<MemInfo>(1).or(Err(libc::EMSGSIZE))? };
-        unsafe {
-            mem_info.as_mut_ptr().write_volatile(untrusted[0]);
-            Ok(mem_info.assume_init())
-        }
+
+        let (_, mem_info) = unsafe { c.read::<MemInfo>() }.or(Err(libc::EMSGSIZE))?;
+
+        Ok(mem_info)
     }
 
     /// Exit the shim with a `status` code
--- a/internal/shim-sev/src/syscall.rs
+++ b/internal/shim-sev/src/syscall.rs
@@ -11,6 +11,7 @@ use crate::payload::{NEXT_BRK_RWLOCK, NEXT_MMAP_RWLOCK};
 use core::convert::TryFrom;
 use core::mem::{size_of, MaybeUninit};
 use core::ops::{Deref, DerefMut};
+use core::ptr::NonNull;
 use primordial::Address;
 use sallyport::request;
 use x86_64::registers::wrfsbase;
@@ -169,32 +170,29 @@ impl Handler {
         hostcall::shim_exit(self.a as _);
     }
 
-    fn _read(&self, fd: usize, data: *mut u8, len: usize) -> Result<usize, libc::c_int> {
+    fn _read(&self, fd: usize, trusted: *mut u8, trusted_len: usize) -> Result<usize, libc::c_int> {
         let mut host_call = HOST_CALL.try_lock().ok_or(libc::EIO)?;
-
-        let trusted: &mut [u8] = unsafe { core::slice::from_raw_parts_mut(data, len) };
 
         let block = host_call.as_mut_block();
 
         let c = block.cursor();
-        let (_, buf) = unsafe { c.alloc::<u8>(trusted.len()).or(Err(libc::EMSGSIZE))? };
+        let (_, buf) = unsafe { c.alloc::<u8>(trusted_len).or(Err(libc::EMSGSIZE))? };
 
         let buf_address = Address::from(buf.as_ptr());
         let phys_unencrypted = ShimPhysUnencryptedAddr::try_from(buf_address).unwrap();
         let host_virt: HostVirtAddr<_> = phys_unencrypted.into();
 
-        block.msg.req = request!(libc::SYS_read => fd, host_virt, len);
+        block.msg.req = request!(libc::SYS_read => fd, host_virt, trusted_len);
         let result = unsafe { host_call.hostcall() };
         let result_len: usize = result.map(|r| r[0].into())?;
 
-        if trusted.len() < result_len {
+        if trusted_len < result_len {
             panic!("syscall read buffer overflow");
         }
 
         let block = host_call.as_mut_block();
         let c = block.cursor();
-        let (_, untrusted) = unsafe { c.alloc(result_len).or(Err(libc::EMSGSIZE))? };
-        trusted[..result_len].copy_from_slice(untrusted);
+        unsafe { c.copy_into_raw_parts(trusted_len, trusted, result_len) }.or(Err(libc::EFAULT))?;
 
         Ok(result_len)
     }
@@ -438,7 +436,7 @@ impl Handler {
     /// Do a ioctl() syscall
     ///
     pub fn ioctl(&self) -> Result<usize, libc::c_int> {
-        match (self.a as i32, self.b as i32) {
+        match (self.a as _, self.b as _) {
             (libc::STDIN_FILENO, libc::TIOCGWINSZ)
             | (libc::STDOUT_FILENO, libc::TIOCGWINSZ)
             | (libc::STDERR_FILENO, libc::TIOCGWINSZ) => {
@@ -538,7 +536,10 @@ impl Handler {
 
     pub fn clock_gettime(&self) -> Result<usize, libc::c_int> {
         let clk_id = self.a;
-        let trusted = self.b as *mut libc::timespec;
+
+        // FIXME: check `trusted`, if in payload space
+        // https://github.com/enarx/enarx-keepldr/issues/78
+        let trusted = NonNull::new(self.b as *mut libc::timespec).ok_or(libc::EFAULT)?;
 
         let mut host_call = HOST_CALL.try_lock().ok_or(libc::EIO)?;
 
@@ -556,10 +557,7 @@ impl Handler {
 
         let block = host_call.as_mut_block();
         let c = block.cursor();
-        let (_, untrusted) = unsafe { c.alloc::<libc::timespec>(1).or(Err(libc::EMSGSIZE))? };
-        unsafe {
-            trusted.write_volatile(untrusted[0]);
-        }
+        unsafe { c.copy_into(trusted) }.or(Err(libc::EMSGSIZE))?;
 
         Ok(result)
     }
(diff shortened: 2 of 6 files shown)


internal/shim-sev/src/syscall.rs
```diff
@@ -537,7 +537,9 @@ impl Handler {
 
     pub fn clock_gettime(&self) -> Result<usize, libc::c_int> {
         let clk_id = self.a;
-        let trusted = self.b as *mut libc::timespec;
+
+        // FIXME: check `trusted`, if in payload space
```
connorkuehl, line 541: Please file an issue for this; we should absolutely be verifying userspace pointers. Once it's filed, put a link to the issue here.

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
+        let slice = unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) };
+
+        // FIXME: should be volatile
+        slice.copy_from_slice(src);
+
         Ok((c, slice))
     }
+
+    /// Copies data into a slice from a slice using self.alloc().
+    #[allow(dead_code)]
```
connorkuehl, line 232: I don't think this is dead code.

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
+        let slice = unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) };
+
+        // FIXME: should be volatile
+        dst.copy_from_slice(slice);
+
+        Ok(c)
+    }
+
+    /// Reads data from the cursor using self.alloc() and advancing the cursor.
+    #[allow(dead_code)]
```
connorkuehl, line 247: I don't think this is dead code either.

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
+        Ok(c)
+    }
+
+    /// Copies data into a an existing buffer from a value using self.alloc() and advancing the cursor.
+    ///
+    /// # Safety
+    /// * The caller has to ensure the destination pointer points to valid memory.
+    /// * The pointer must be properly aligned.
+
+    #[allow(dead_code)]
```
connorkuehl, line 277: Doesn't appear to be dead.

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
+        let mut dst = MaybeUninit::<T>::uninit();
+
+        unsafe {
+            dst.as_mut_ptr().write(src.read_volatile());
+        }
+        Ok((c, unsafe { dst.assume_init() }))
+    }
+
+    /// Writes data into the cursor using self.alloc() and advancing the cursor.
+    #[allow(dead_code)]
```
connorkuehl, line 261: I don't think this is dead either.

connorkuehl (changes requested):
Have you seen the [volatile](https://crates.io/crates/volatile) crate? Could it help address your `FIXME`s regarding volatile slice read/writes?

internal/shim-sgx/src/handler.rs
```diff
@@ -209,9 +210,9 @@ impl<'a> Handler<'a> {
 
         let c = self.block.cursor();
         let trusted: &[u8] = unsafe { self.aex.gpr.rsi.into_slice(self.aex.gpr.rdx) };
-        let (_, untrusted) = c.copy_slice(trusted).or(Err(libc::EMSGSIZE))?;
+        let (_, untrusted) = c.copy_from_slice(trusted).or(Err(libc::EMSGSIZE))?;
 
-        let req = request!(libc::SYS_write => self.aex.gpr.rdi, untrusted, untrusted.len());
+        let req = request!(libc::SYS_write => self.aex.gpr.rdi, untrusted, trusted.len());
```
lkatalin, line 215: Similar to the above, why this change?

lkatalin (commented):
This looks pretty good; I had questions about why the PR changes the definitions of `count` bytes to read / write. Also just for curiosity's sake some questions around `[MaybeUninit<T>] as [T]` vs. `assume_init()`, but I can look into that more as it's just for my own understanding.

haraldh:
> Have you seen the [volatile](https://crates.io/crates/volatile) crate? Could it help address your `FIXME`s regarding volatile slice read/writes?

Yes, that's why I am referring in the commit message with
Related: #59

src/sallyport/mod.rs
haraldh: This is because of cargo and publishing `enarx-keepldr` on crates.io.
We therefore do not have `sallyport` as a dependency in `enarx-keepldr`, but `src/sallyport/mod.rs` and for the shims the `internal/sallyport` crate.
Although `src/sallyport/mod.rs` looks like a library crate, it is not and the compiler and clippy will warn, if any function is not used in the `enarx-keepldr`, which are most of it.

src/sallyport/mod.rs
haraldh: see above

src/sallyport/mod.rs
haraldh: see above

src/sallyport/mod.rs
haraldh: see above

internal/shim-sev/src/hostcall.rs
```diff
@@ -92,13 +91,13 @@ impl<'a> HostCall<'a> {
...
         let cursor = self.0.cursor();
-        let (_, buf) = cursor.copy_slice(bytes).or(Err(libc::EMSGSIZE))?;
+        let (_, buf) = cursor.copy_from_slice(bytes).or(Err(libc::EMSGSIZE))?;
 
         let buf_address = Address::from(buf.as_ptr());
         let phys_unencrypted = ShimPhysUnencryptedAddr::try_from(buf_address).unwrap();
         let host_virt: HostVirtAddr<_> = phys_unencrypted.into();
 
-        self.0.msg.req = request!(libc::SYS_write => fd, host_virt, buf.len());
+        self.0.msg.req = request!(libc::SYS_write => fd, host_virt, bytes.len());
```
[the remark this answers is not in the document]
haraldh: `buf.len()` and `bytes.len()` are equal, because using `Cursor::copy_from_slice()` uses `Cursor::alloc()`, which allocates `bytes.len()` with or fails.
```
        let (_, buf) = cursor.copy_from_slice(bytes).or(Err(libc::EMSGSIZE))?;
```

Anyway... I changed it back.

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
-    pub fn copy_slice<T: Copy>(
+    pub fn copy_from_slice<T: 'a + Copy>(
         self,
-        value: &[T],
+        src: &[T],
     ) -> core::result::Result<(Cursor<'a>, &'a mut [T]), ()> {
-        let (c, slice) = unsafe { self.alloc(value.len())? };
-        slice.copy_from_slice(value);
+        let (c, slice) = unsafe { self.alloc::<T>(src.len())? };
+        let slice = unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) };
```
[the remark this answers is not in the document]
haraldh: See https://doc.rust-lang.org/core/mem/union.MaybeUninit.html#initializing-an-array-element-by-element
which uses:
```rust
unsafe { mem::transmute::<_, [Vec<u32>; 1000]>(data) }
```
but clippy suggested using instead:
```rust
unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) };
```

There is no `slice.assume_init()` because `slice` is `&[MaybeUninit<T>]` and not `MaybeUnit<[T]>`

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
+        Ok(c)
+    }
+
+    /// Reads data from the cursor using self.alloc() and advancing the cursor.
+    #[allow(dead_code)]
+    pub fn read_volatile<T: 'a + Copy>(self) -> core::result::Result<(Cursor<'a>, T), ()> {
+        let (c, slice) = unsafe { self.alloc::<T>(1)? };
+
+        let src = slice[0].as_ptr();
+        let mut dst = MaybeUninit::<T>::uninit();
```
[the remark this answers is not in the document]
haraldh: `uninit()` is basically `new()` for `MaybeUninit<T>`. You cannot create `MaybeUninit<T>` yourself. You have to use either `uninit()` or `zeroed()` or something else. `MaybeUninit::<T>::uninit()` is basically `malloc(sizeof(T))`, where the contents of the allocated memory is undefined, which does not matter here, because we fill it with ` dst.as_mut_ptr().write()` right afterwards.

internal/shim-sgx/src/handler.rs
```diff
@@ -40,9 +41,9 @@ impl<'a> Write for Handler<'a> {
         }
 
         let c = self.block.cursor();
-        let (_, untrusted) = c.copy_slice(s.as_bytes()).or(Err(core::fmt::Error))?;
+        let (_, untrusted) = c.copy_from_slice(s.as_bytes()).or(Err(core::fmt::Error))?;
 
-        let req = request!(libc::SYS_write => libc::STDERR_FILENO, untrusted, untrusted.len());
+        let req = request!(libc::SYS_write => libc::STDERR_FILENO, untrusted, s.len());
```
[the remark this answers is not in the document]
haraldh: both must be the same length... anyway.. this was a leftover from old code... switched back.

internal/shim-sgx/src/handler.rs
haraldh: both must be the same length... anyway.. this was a leftover from old code... switched back.

internal/shim-sev/src/syscall.rs
haraldh: done

haraldh: https://github.com/enarx/enarx-keepldr/issues/78

src/sallyport/mod.rs
```diff
@@ -195,32 +197,95 @@ impl<'a> Cursor<'a> {
...
     }
+
+    /// Copies data into a slice from a slice using self.alloc().
+    #[allow(dead_code)]
+    pub fn copy_into_slice<T: 'a + Copy>(
+        self,
+        dst: &mut [T],
+    ) -> core::result::Result<Cursor<'a>, ()> {
+        let (c, slice) = unsafe { self.alloc::<T>(dst.len())? };
+        let slice = unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) };
```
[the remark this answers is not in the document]
haraldh: You are right. Let me turn that function into `unsafe` with the caller _ensuring_ the `Cursor` contains valid data.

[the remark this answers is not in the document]
haraldh: Oops, no remove the `unsafe` and changed `dst` to be `dst: &mut [MaybeUninit<T>]`.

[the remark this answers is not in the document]
haraldh: Thinking more about it, I changed it to:
```rust
    pub fn copy_into_raw_parts<T: 'a + Copy>(self, dst: *mut MaybeUninit<T>, len: usize) 
```
in preparation for [`intrinsics::volatile_copy_nonoverlapping_memory`](https://doc.rust-lang.org/core/intrinsics/fn.volatile_copy_nonoverlapping_memory.html)

src/sallyport/mod.rs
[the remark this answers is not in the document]
haraldh: completely overhauled the whole thing

src/sallyport/mod.rs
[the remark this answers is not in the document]
haraldh: scratch the above, completely overhauled the whole thing again

haraldh:
Phew... tested the whole thing with `miri` also.

npmccallum (changes requested):
Please split the `MaybeUninit` into a separate commit. This will make it easier to review the `volatile` change independently.

haraldh:
Split into several commits.

The use of the compiler memory fence made the volatile functions obsolete.

The use of the raw pointers, made the use of `MaybeUnit` mostly obsolete.

src/sallyport/mod.rs
```diff
@@ -174,7 +176,9 @@ impl Default for Block {
 impl Block {
     /// Returns the capacity of `Block.buf`
     pub const fn buf_capacity() -> usize {
-        512 * Page::size() - size_of::<Message>()
+        // FIXME: https://github.com/enarx/enarx-keepldr/issues/23
+        let page_num = if cfg!(test) { 1 } else { 512 };
+        page_num * Page::size() - size_of::<Message>()
```
connorkuehl, line 181: Why only for test builds?

Also, this is enormous, and we won't be able to construct the keep due to the 2MiB footprint assert in BootInfo::calculate.

512 * 4096 = 2097152 = 2 MiB (per CPU).

Which means that a single CPU configuration for the keep already depletes the quota enforced by BootInfo::calculate leaving no room for the shim or code.

This probably didn't come up in testing because the `kvm` backend mistakenly assumes that `size_of::<sallyport::Block>() == size_of::<Page>()`.

The following correction to the loader will uncover the assert failure.

```diff
diff --git a/src/backend/kvm/vm/builder.rs b/src/backend/kvm/vm/builder.rs
index 95e1001..8a40b97 100644
--- a/src/backend/kvm/vm/builder.rs
+++ b/src/backend/kvm/vm/builder.rs
@@ -113,7 +113,7 @@ impl<T: Hook> Builder<T> {
         code_size: Span<usize>,
     ) -> Result<BootInfo> {
         let setup_size = Page::size()
-            + (Page::size() * cfg.n_syscall_pages.get())
+            + (crate::sallyport::Block::buf_capacity() * cfg.n_syscall_pages.get())
             + size_of::<PageTable>()
             + size_of::<PageTable>();
```

I think the only other bugs with the loader WRT sallyport size will be `arch_specific_setup` since it writes a zeroed page to each sallyport block.

connorkuehl (changes requested):

src/sallyport/mod.rs
haraldh: Huh? I didn't change anything for the non-test case:

It was:
```diff
-        512 * Page::size() - size_of::<Message>()
```

connorkuehl: Yeah, but don't we want them to be the same size for test and non-test?

haraldh: We can't test big `Block` sizes on bare metal, because the `Block` is created on the stack.

haraldh: And, of course I will fix the whole thing with a follow-up patch. This is why the `FIXME` with a pointer to the issue is still in there.

connorkuehl (approved):

npmccallum (approved):

PR #94, opened 2020-10-21
Add getuid syscall test

### LINKED ISSUE #93: Implement getuid syscall test
Execution issue for the getuid portion of https://github.com/enarx/enarx-keepldr/issues/9

ambaxter:
Closes #93.

Signed-off-by: Adam Baxter <ambaxter@users.noreply.github.com>

### DIFF
--- a/tests/bin/getuid.c
+++ b/tests/bin/getuid.c
@@ -1,0 +1,8 @@
+#include "libc.h"
+
+int main(void) {
+    if (!is_enarx()) {
+        return 0;
+    }
+    return getuid() != 1000;
+}
--- a/tests/bin/libc.h
+++ b/tests/bin/libc.h
@@ -3,6 +3,7 @@
 #include <unistd.h>
 #include <errno.h>
 #include <time.h>
+#include <sys/types.h>
 
 int *__errno_location(void) {
     static int errnum = 0;
@@ -132,5 +133,15 @@ ssize_t get_att(void *nonce, size_t nonce_len, void *buf, size_t buf_len, size_t
     }
 
     *technology = tech;
+    return rax;
+}
+
+uid_t getuid() {
+    uid_t rax;
+    asm(
+        "syscall"
+        : "=a" (rax)
+        : "a" (SYS_getuid)
+    );
     return rax;
 }
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -169,3 +169,8 @@ fn echo() {
 fn get_att() {
     run_test("get_att", 0, None, None, None);
 }
+
+#[test]
+fn getuid() {
+    run_test("getuid", 0, None, None, None);
+}


tests/bin/getuid.c
```diff
@@ -0,0 +1,7 @@
+#include "libc.h"
+
+int main(void) {
+    uid_t uid = 0;
+    uid = getuid();
+    return uid == 0;
+}
```
npmccallum: Newline at the end please.

tests/bin/libc.h
```diff
@@ -73,3 +74,13 @@ int is_enarx() {
...
+
+uid_t getuid() {
+    uid_t rax;
+    asm(
+        "syscall" 
+        : "=a" (rax)
+        : "a" (SYS_getuid)
+    );
+    return rax;
+}
```
npmccallum: Newline at the end.

npmccallum (changes requested):
This test fails because it appears that SGX and SEV have differing implementations of `getuid()`. SEV has no implementation of `getuid()`. SGX [returns the actual UID](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sgx/src/handler.rs#L182). We need to agree on a shared behavior. Alternatively, we could remove support for getuid altogether. @haraldh @connorkuehl Thoughts?

connorkuehl:
I think SEV should be updated to proxy the `getuid()` call as well.

haraldh:
> This test fails because it appears that SGX and SEV have differing implementations of `getuid()`. SEV has no implementation of `getuid()`. SGX [returns the actual UID](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sgx/src/handler.rs#L182). We need to agree on a shared behavior. Alternatively, we could remove support for getuid altogether. @haraldh @connorkuehl Thoughts?

just return a fake uid, like `1000`

ambaxter:
> This test fails because it appears that SGX and SEV have differing implementations of `getuid()`. SEV has no implementation of `getuid()`. SGX [returns the actual UID](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sgx/src/handler.rs#L182). We need to agree on a shared behavior. Alternatively, we could remove support for getuid altogether. @haraldh @connorkuehl Thoughts?

Is this something I should change as part of this PR or is this a different issue?

haraldh:
> > This test fails because it appears that SGX and SEV have differing implementations of `getuid()`. SEV has no implementation of `getuid()`. SGX [returns the actual UID](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sgx/src/handler.rs#L182). We need to agree on a shared behavior. Alternatively, we could remove support for getuid altogether. @haraldh @connorkuehl Thoughts?
> 
> Is this something I should change as part of this PR or is this a different issue?

sure! then the PR tests succeed :smile:

npmccallum:
This test now depends on PR #98.

tests/bin/getuid.c
```diff
@@ -0,0 +1,10 @@
+#include "libc.h"
+
+int main(void) {
+    uid_t uid = 0xFFFFFFFF;
+    uid = getuid();
+    if(is_enarx()) {
+        return uid != 1000;
+    }
+    return uid == 0xFFFFFFFF;
```
npmccallum, line 9: I think this workflow is simpler. Otherwise, everything looks good.

```suggestion
    if (!is_enarx())
        return 0;

    return getuid() != 1000;
```

npmccallum (changes requested):

tests/bin/getuid.c
ambaxter: Done

tests/bin/getuid.c
```diff
@@ -0,0 +1,8 @@
+#include "libc.h"
+
+int main(void) {
+    if(!is_enarx()) {
+        return 0;
+    }
```
npmccallum, line 6:
```suggestion
    if (!is_enarx())
        return 0;
```

npmccallum (changes requested):

tests/bin/getuid.c
haraldh: In general, I prefer the original version, because it's better for patches.

haraldh (approved):

tests/bin/getuid.c
npmccallum: I'll accept the braces. I require a space after `if`, however.

ambaxter: Done.

npmccallum (approved):

haraldh (approved):

npmccallum (changes requested):
This needs to be rebased.

npmccallum (approved):

connorkuehl (approved):

PR #95, opened 2020-10-21
Preparatory work for management by enarx-keepmgr

MikeCamel:
Currently doesn't actually spawn a keep.
Needs to create Unix socket and communicate with enarx-keepmgr.

Signed-off-by: MikeCamel <mike@p2ptrust.org>

### DIFF
--- a/external/enarx-keep-kvm@.service
+++ b/external/enarx-keep-kvm@.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Enarx Keep-Loader (KVM)
+Documentation=https://github.com/enarx/
+
+[Service]
+Type=simple
+Environment="ENARX_BACKEND=kvm"
+ExecStart=/home/mike/programming/enarx-keepldr/target/debug/enarx-keepldr kuuid I%
+
+UMask=0066
+
+StandardOutput=file:/home/mike/enarx-keepldr-%i.log
+StandardError=file:/home/mike/enarx-keepldr-%i.err
+
--- a/external/enarx-keep-kvm@.service
+++ b/external/enarx-keep-kvm@.service
@@ -9,6 +9,6 @@ ExecStart=/home/mike/programming/enarx-keepldr/target/debug/enarx-keepldr kuuid 
 
 UMask=0066
 
-StandardOutput=file:/home/mike/enarx-keepldr-%i.log
-StandardError=file:/home/mike/enarx-keepldr-%i.err
+StandardOutput=journal+console
+StandardError=journal+console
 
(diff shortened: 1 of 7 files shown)


external/enarx-keep-kvm@.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Enarx Keep-Loader (KVM)
+Documentation=https://github.com/enarx/
+
+[Service]
+Type=simple
+Environment="ENARX_BACKEND=kvm"
+ExecStart=/home/mike/programming/enarx-keepldr/target/debug/enarx-keepldr kuuid I%
```
haraldh, line 8: Your personal paths don't belong in the repo :-1:

src/main.rs
```diff
@@ -45,6 +45,10 @@
 //!     $ musl-gcc -static-pie -fPIC -o test test.c
 //!     $ target/debug/enarx-keepldr exec ./test
 //!     Hello World!
+//!
+//!     Or copy (or link) the files in external/*.service to ~/.config/systemd/user/
+//!     $ systemctl --user daemon-reload
+//!     $ systemctl --user start enarx-keep-nil@[Uuid].service
```
haraldh, line 51: I think this should not run in systemd `--user` mode, but system wide.

haraldh (changes requested):

external/enarx-keep-kvm@.service
MikeCamel: Fair - what's the canonical place for this sort of thing?

src/main.rs
MikeCamel: I agree, but there's a balance between allowing multiple tests on a shared system and running as a particular user. I'm not sure what the correct approach is, tbh.

external/enarx-keep-kvm@.service
haraldh: Well, define a placeholder and document, that the path has to be fixed

src/main.rs
haraldh: Then, document: "for testing purposes […]"

external/enarx-keep-kvm@.service
haraldh: Could be  `@BIN@` `@LOG@` ...

haraldh (changes requested):
still failing

MikeCamel:
> still failing

I've fixed the README.md test - now failing on a sallyport issue.   Not me...

haraldh:
> > still failing
> 
> I've fixed the README.md test - now failing on a sallyport issue. Not me...

yay... the benefits of cargo deny all and using cargo nightly... new spurious CI failures  /s

npmccallum:
After a conversation, the `kuuid` command is unnecessary. Closing.

PR #99, opened 2020-10-22
shim-sev: refactor, simplify, beautify the syscalls

haraldh:
* shim-sev:syscall: Add the C arguments to the syscalls
    
    This patch adds the C arguments to the syscalls to match the signature
    of the C functions. This improves recognition and readability.
    

* shim-sev:syscall: Remove `Handler`
    
    Remove the now obsolete `Handler` struct, which saves one indentation
    level and provides the posssibility to move those functions in separate
    modules, like `io`, `mem`, etc.

* shim-sev:syscall: Fix write syscall
    
    Don't use `host_call.write()` but rather do a complete syscall.
    `host_call.write()` should be used kernel internal only.

* shim-sev:syscall: Factor out `blk_to_host_virt`
    
    The `blk_to_host_virt()` function factors out the often used conversion from
    a `sallyport:Block` buffer address to the host virtual address.

### DIFF
--- a/internal/shim-sev/src/hostcall.rs
+++ b/internal/shim-sev/src/hostcall.rs
@@ -98,7 +98,7 @@ impl<'a> HostCall<'a> {
     /// # Safety
     ///
     /// The parameters returned can't be trusted.
-    pub unsafe fn write(&mut self, fd: usize, bytes: &[u8]) -> sallyport::Result {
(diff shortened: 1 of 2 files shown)


npmccallum (approved):
There are still a lot of casts hanging around even after the adoption of `Register`. At some point we should address these.

connorkuehl (approved):

PR #100, opened 2020-10-22
Create a test for the getatt() syscall that should return ENOSYS

### LINKED ISSUE #89: Write a test for the getatt() pseudo-syscall
This pseudo-syscall `getatt()` is from code layer to the shim layer for multiple technologies (see #31 ). At first, the test should pass if the return value is `ENOSYS` (because this syscall is not yet implemented). A return value of `0` when the backend is SGX will also allow the test to pass.

lkatalin:
Closes #89 .

### DIFF
--- a/tests/bin/get_att.c
+++ b/tests/bin/get_att.c
@@ -1,0 +1,15 @@
+#include "libc.h"
+#include <errno.h>
+
+int main(void) {
+    // TODO: Good buffer length?
+    unsigned char nonce[512];
+    unsigned char buf[512];
+    size_t technology;
+
+    ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
+
+    // TODO: Currently this should return ENOSYS. However, this will change
+    // as get_att() is implemented for each technology.
+    return !(size == -1 && errno == ENOSYS);
+}
--- a/tests/bin/libc.h
+++ b/tests/bin/libc.h
@@ -113,3 +113,24 @@ int is_enarx() {
         default: return 0;
     }
 }
+
+ssize_t get_att(void *nonce, size_t nonce_len, void *buf, size_t buf_len, size_t *technology) {
+    ssize_t rax;
+    ssize_t tech;
+    register size_t r10 __asm__("r10") = buf_len;
+
+    asm(
+        "syscall"
+        : "=a" (rax), "=d" (tech)
+        : "a" (0xEA01), "D" (nonce), "S" (nonce_len), "d" (buf), "r" (r10)
+        : "%rcx", "%r11"
+    );
+
+    if (rax < 0) {
+        errno = -rax;
+        return -1;
+    }
+
+    *technology = tech;
+    return rax;
+}
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -164,3 +164,8 @@ fn echo() {
     }
     run_test("echo", 0, input.as_slice(), input.as_slice(), None);
 }
+
+#[test]
+fn get_att() {
+    run_test("get_att", 0, None, None, None);
+}


tests/bin/libc.h
```diff
@@ -73,3 +73,18 @@ int is_enarx() {
...
     }
 }
+
+ssize_t get_att(void *nonce, size_t nonce_len, void *buf, size_t buf_len, size_t *technology) {
+    ssize_t rax;
+    ssize_t tech;
+    register size_t r10 __asm__("r10") = buf_len;
+
+    asm(
+        "syscall"
```
npmccallum, line 83: Indent.

tests/bin/libc.h
```diff
@@ -73,3 +73,18 @@ int is_enarx() {
...
+	: "=a" (rax), "=d" (tech)
+	: "a" (0xEA01), "D" (nonce), "S" (nonce_len), "d" (buf), "r" (r10)
+    );
```
npmccallum, line 86: I recently learned that the `syscall` instruction clobbers `rcx` and (I think...) `r11`. That could be a source of the bug here.

tests/bin/get_att.c
```diff
@@ -0,0 +1,18 @@
...
+
+int main(void) {
+    // TODO: Good buffer length?
+    unsigned char nonce[512];
+    unsigned char buf[512];
+    size_t technology;
+
+    void* nonce_ptr = &nonce;
+    void* buf_ptr = &buf;
+    size_t* tech_ptr = &technology;
```
npmccallum, line 11: No need for intermediate variables here.

npmccallum: Also note, `nonce` and `buf` are already pointers to buffers. That's how C arrays work.

tests/bin/get_att.c
```diff
@@ -0,0 +1,18 @@
...
+
+    ssize_t rax = get_att(nonce_ptr, sizeof(nonce), buf_ptr, sizeof(buf), tech_ptr);
+
+    // TODO: Currently this should return ENOSYS. However, this will change
+    // as get_att() is implemented for each technology.
+    return rax != ENOSYS;
```
npmccallum, line 17: `return rax != -ENOSYS;`

Notice the negative. The Linux syscall ABI for x86_64 specifies that the kernel returns values in `rax` and `rdx` unless an error occurs. Then it returns a negative errno in `rax`.

tests/bin/libc.h
npmccallum: I have added fixes for the existing syscalls in #101.

tests/bin/get_att.c
npmccallum: This needs to be updated to check `errno`.

tests/bin/libc.h
lkatalin: Good to know!

tests/bin/get_att.c
lkatalin: You make a very good point.

tests/bin/libc.h
lkatalin: @npmccallum How do you tend to check that your indentation is correct (here and in general)? I tend to mess up on this.

npmccallum: Basically, you just need to configure your editor for automatic indenting. So it depends on your editor.

npmccallum: Technically, `"syscall"` is the correct indent. It is the lines below that are incorrect.

tests/bin/get_att.c
npmccallum: Drop `tech_ptr`. It isn't needed.

tests/bin/get_att.c
```diff
@@ -0,0 +1,16 @@
...
+#include <errno.h>
+
+int main(void) {
+    // TODO: Good buffer length?
+    unsigned char nonce[512];
+    unsigned char buf[512];
+    size_t technology;
+    size_t* tech_ptr = &technology;
+
+    ssize_t rax = get_att(nonce, sizeof(nonce), buf, sizeof(buf), tech_ptr);
```
npmccallum, line 11:
```suggestion
    ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
```

tests/bin/get_att.c
```diff
@@ -0,0 +1,16 @@
...
+
+    // TODO: Currently this should return ENOSYS. However, this will change
+    // as get_att() is implemented for each technology.
+    return errno != ENOSYS;
```
npmccallum, line 15: You also need to check the return value of `get_att()` (which should be `-1` for `errno` to be set).

tests/bin/libc.h
```diff
@@ -113,3 +113,24 @@ int is_enarx() {
...
+ssize_t get_att(void *nonce, size_t nonce_len, void *buf, size_t buf_len, size_t *technology) {
+    ssize_t rax;
+    ssize_t tech;
+    register size_t r10 __asm__("r10") = buf_len;
+
+    asm(
+        "syscall"
+	: "=a" (rax), "=d" (tech)
+	: "a" (0xEA01), "D" (nonce), "S" (nonce_len), "d" (buf), "r" (r10)
+	: "%rcx", "%r11"
```
npmccallum, line 126: Please fix this indent.

npmccallum (changes requested):

tests/bin/libc.h
lkatalin: Ah, I figured out why it looks the same to me. It was a tabs vs. spaces thing again. I'll look into how to make vim do some of this for me.

tests/bin/get_att.c
```diff
@@ -0,0 +1,15 @@
...
+    // TODO: Good buffer length?
+    unsigned char nonce[512];
+    unsigned char buf[512];
+    size_t technology;
+
+    ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
+
+    // TODO: Currently this should return ENOSYS. However, this will change
+    // as get_att() is implemented for each technology.
+    return (size != -1 || errno != ENOSYS);
```
npmccallum, line 14: I find this logic hard to grok without thinking. I think we should do this instead, since it is much more obvious.

```suggestion
    return !(size == -1 && errno == ENOSYS);
```

The logic inside the parentheses expresses our success condition. The exclamation point outside the parentheses inverts the success condition into an exit status.

npmccallum (changes requested):

npmccallum (approved):

connorkuehl (approved):

PR #101, opened 2020-10-22
Two fixes for our test infrastructure libc implementation

npmccallum:
I think we should be POSIX compliant after these fixes.


lkatalin (approved):
This is both useful and interesting.

haraldh (approved):
LGTM

connorkuehl (approved):

PR #102, opened 2020-10-22
-static-pie was not an option for musl, but -static and -pie were. Is this just on my system (4.15.0-122-generic #124-Ubuntu SMP Thu Oct 15 13:03:05 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux)

frumioj:

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ technologies such as Intel SGX or AMD SEV.
     }
     EOF
 
-    $ musl-gcc -static-pie -fPIC -o test test.c
+    $ musl-gcc -static -pie -fPIC -o test test.c
     $ target/debug/enarx-keepldr exec ./test
     Hello World!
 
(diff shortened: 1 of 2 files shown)


connorkuehl:
@frumioj Based on your kernel version I'm guessing you're running Ubuntu Bionic or Xenial (with an edge kernel). I'm guessing the `gcc` version in both of those releases are too old to have the compiler options listed here in the documentation.

Could you share the output of `gcc --version`?

frumioj:
I added gcc-8 to my machine, and recompiled with that. It is now working. I will close this PR. Thanks!

frumioj:
Since Ubuntu 18.04 (at least) comes with gcc 7.4, and it seems the minimum required version is gcc 8.4 (that's what I get for sudo apt install gcc-8) I would recommend adding a note to the README file.

frumioj:
Actually scratch that. It looks like you need a minimum of gcc-9 with the backported mtune=znver2 switch for musl-gcc. I still recommend a note on the README :)

haraldh:
That reminds me, that `mtune=znver2` can be removed, if it doesn't work for all.

PR #103, opened 2020-10-22
Remove __thread storage class from errnum

connorkuehl:
Otherwise we segfault when we access it for some reason...

### DIFF
--- a/tests/bin/libc.h
+++ b/tests/bin/libc.h
@@ -5,7 +5,7 @@
 #include <time.h>
 
 int *__errno_location(void) {
-    static __thread int errnum = 0;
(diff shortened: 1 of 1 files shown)


npmccallum (approved):

lkatalin (approved):

haraldh:
> Otherwise we segfault when we access it for some reason...
> 
> Signed-off-by: Connor Kuehl [ckuehl@redhat.com](mailto:ckuehl@redhat.com)

My guess is, that the thread local `fsbase` pointing to the thread local storage setup/config is not setup with `arch_prctl(ARCH_SET_FS, …)`, because `crt*` is missing.

PR #108, opened 2020-10-23
Add close syscall test

### LINKED ISSUE #107: Implement close syscall test
Execution issue for the close portion of #9

greyspectrum:
Closes #107

### DIFF
--- a/tests/bin/close.c
+++ b/tests/bin/close.c
@@ -1,0 +1,6 @@
+#include "libc.h"
+
+int main(void) {
+    ssize_t ret = close(STDIN_FILENO);
+    return ret != 0;
+}
--- a/internal/syscall/src/lib.rs
+++ b/internal/syscall/src/lib.rs
@@ -167,6 +167,7 @@ pub trait SyscallHandler: AddressValidator + Sized {
             libc::SYS_getgid => self.getgid(),
             libc::SYS_geteuid => self.geteuid(),
             libc::SYS_getegid => self.getegid(),
+            libc::SYS_close => self.close(a.try_into().map_err(|_| libc::EINVAL)?),
 
             SYS_GETATT => self.get_attestation(),
 
--- a/tests/bin/libc.h
+++ b/tests/bin/libc.h
@@ -181,3 +181,21 @@ gid_t getegid() {
     );
     return rax;
 }
+
+int close(int fd) {
+    ssize_t rax;
+
+    asm(
+        "syscall"
+        : "=a" (rax)
+        : "a" (SYS_close), "D" (fd)
+        : "%rcx", "%r11"
+    );
+
+    if (rax < 0) {
+        errno = -rax;
+        return -1;
+    }
+
+    return rax;
+}
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -132,6 +132,12 @@ fn clock_gettime() {
 
 #[test]
 #[serial]
+fn close() {
+    run_test("close", 0, None, None, None);
+}
+
+#[test]
+#[serial]
 fn write_stdout() {
     run_test("write_stdout", 0, None, &b"hi\n"[..], None);
 }


tests/bin/close.c
```diff
@@ -0,0 +1,5 @@
+#include "libc.h"
+
+int main(void) {
+    close(STDIN_FILENO);
```
haraldh, line 4: To really test this, you should check the return value of `close()` and maybe return `errno` in this case.

haraldh (changes requested):

tests/bin/close.c
connorkuehl: I'm wondering if the return value from `close` is just implicitly held in `rax` and despite not having a return statement, it is implicitly returned to the system. I think perhaps this should be made explicit.

greyspectrum: I added an expression to check the return value of `close()`, which should evaluate to false (so zero) if close() returns zero. However, the test now fails and I'm not sure why. I see that close() is [implemented as a proxied call to the host](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sgx/src/handler.rs#L543), so there's nothing unusual there. `close()` should return zero on success, so perhaps it's not possible to call `close()` on `STDIN`? Otherwise I'm not quite sure where I went wrong, unfortunately.

I'm not sure how to return `errno`, since the test is expected to return one or zero, but `errno` has multiple error return values.

lkatalin: It seems `close()` is not yet implemented in the SEV backend, so this test will fail until we implement it there. It should work with `cargo test --features=backend-sgx`. However, it won't make sense to merge this until `close()` is included in SEV.

greyspectrum: I implemented close() for the SEV backend, [here](https://github.com/enarx/enarx-keepldr/pull/121).

However, I noticed that if I run `cargo test --features=backend-sgx` against the `close` branch, here, the test for close still fails. So, there may be something else going on and merging #121 may not fix it.

tests/bin/close.c
```diff
@@ -0,0 +1,6 @@
+#include "libc.h"
+
+int main(void) {
+    ssize_t rax = close(STDIN_FILENO);
```
haraldh, line 4: that should probably be `ret` instead of `rax` :smiley:

haraldh (changes requested):
Needs rebasing to current master, which will make the tests most likely pass.

haraldh:
Oops.. it still needs one PR.. hold on

haraldh:
Include this in your PR:
```diff
diff --git a/internal/syscall/src/lib.rs b/internal/syscall/src/lib.rs
index 41f7f49..7f0cfe4 100644
--- a/internal/syscall/src/lib.rs
+++ b/internal/syscall/src/lib.rs
@@ -167,6 +167,7 @@ pub trait SyscallHandler: AddressValidator + Sized {
             libc::SYS_getgid => self.getgid(),
             libc::SYS_geteuid => self.geteuid(),
             libc::SYS_getegid => self.getegid(),
+            libc::SYS_close => self.close(a.try_into().map_err(|_| libc::EINVAL)?),
 
             SYS_GETATT => self.get_attestation(),
 ```

haraldh:
You forgot `Signed-off-by: greyspectrum <hello@greyspectrum.io>` in the commit message

greyspectrum:
> You forgot `Signed-off-by: greyspectrum <hello@greyspectrum.io>` in the commit message

Fixed.

greyspectrum:
It looks like two tests are failing for AMD SEV. I have an Intel machine, so tests pass locally.

connorkuehl:
> It looks like two tests are failing for AMD SEV. I have an Intel machine, so tests pass locally.

Yeah, our expanded integration test coverage has uncovered a bottleneck with our SEV VM launch process. I have patches to ameliorate this for now: https://github.com/enarx/enarx-keepldr/pull/158

lkatalin (approved):

haraldh (approved):

haraldh:
Congratulations 👏 🥳

PR #114, opened 2020-10-26
Code => shim stub for getatt() syscall in SGX

### LINKED ISSUE #90: Implement stub for getatt() pseudo-syscall (code => shim) in SGX
When successful, the `technology` ID output buffer should return `2` for SGX as described in #31 .

lkatalin:
Closes #90

### DIFF
--- a/internal/shim-sev/src/syscall.rs
+++ b/internal/shim-sev/src/syscall.rs
@@ -21,6 +21,10 @@ use x86_64::{align_up, VirtAddr};
 const FAKE_UID: usize = 1000;
 const FAKE_GID: usize = 1000;
 
+// See https://github.com/enarx/enarx-keepldr/issues/31
+const SEV_TECH: usize = 1;
+const SYS_GETATT: usize = 0xEA01;
+
 #[repr(C)]
 struct X8664DoubleReturn {
     rax: u64,
@@ -178,8 +182,10 @@ extern "sysv64" fn syscall_rust(
     if nr >= 0xEA00 {
         // Enarx syscalls
 
-        // Fixme
-        let ret: sallyport::Result = Err(libc::ENOSYS);
+        let ret: sallyport::Result = match nr {
+            SYS_GETATT => Ok([0.into(), SEV_TECH.into()]),
+            _ => Err(libc::ENOSYS),
+        };
 
         match ret {
             Ok([rax, rdx]) => X8664DoubleReturn {
--- a/internal/shim-sgx/src/handler.rs
+++ b/internal/shim-sgx/src/handler.rs
@@ -17,6 +17,9 @@ pub const TRACE: bool = false;
 const FAKE_UID: usize = 1000;
 const FAKE_GID: usize = 1000;
 
+// See https://github.com/enarx/enarx-keepldr/issues/31
+const SGX_TECH: usize = 2;
+
 // arch_prctl syscalls not available in the libc crate as of version 0.2.69
 const ARCH_SET_GS: usize = 0x1001;
 const ARCH_SET_FS: usize = 0x1002;
@@ -572,5 +575,13 @@ impl<'a> Handler<'a> {
         unsafe { c.copy_into_raw_parts(nfds as _, trusted, nfds as _) }.or(Err(libc::EMSGSIZE))?;
 
         Ok(result)
+    }
+
+    // Stub for get_attestation() pseudo syscall
+    // See: https://github.com/enarx/enarx-keepldr/issues/31
+    pub fn get_att(&mut self) -> sallyport::Result {
+        self.trace("get_att", 0);
+        let rep: sallyport::Reply = Ok([0.into(), SGX_TECH.into()]).into();
+        sallyport::Result::from(rep)
     }
 }
--- a/tests/bin/get_att.c
+++ b/tests/bin/get_att.c
@@ -1,6 +1,12 @@
 #include "libc.h"
 #include <errno.h>
 
+typedef enum {
+    NO_KEEP,
+    SEV,
+    SGX,
+} tech;
+
 int main(void) {
     // TODO: Good buffer length?
     unsigned char nonce[512];
@@ -9,7 +15,19 @@ int main(void) {
 
     ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
 
-    // TODO: Currently this should return ENOSYS. However, this will change
-    // as get_att() is implemented for each technology.
-    return !(size == -1 && errno == ENOSYS);
+    if (size >= 0) {
+	switch (technology) {
+	case NO_KEEP:
+	case SEV:
+	case SGX:
+	    return 0;
+	default: return 1;
+	}
+    }
+
+    else if (size == -1)
+	return !(errno == ENOSYS);
+
+    else
+	return 1;
 }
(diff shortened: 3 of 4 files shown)


src/sallyport/mod.rs
```diff
@@ -90,8 +90,10 @@ impl Request {
 #[repr(C)]
 #[derive(Copy, Clone, Default, PartialEq, Debug)]
 pub struct Reply {
-    ret: [Register<usize>; 2],
-    err: Register<usize>,
+    /// Return values in registers rax:rdx
+    pub ret: [Register<usize>; 2],
+    /// Any error code associated with the Reply
+    pub err: Register<usize>,
```
npmccallum, line 96: Don't make these public.

internal/shim-sgx/src/handler.rs
```diff
@@ -564,4 +567,17 @@ impl<'a> Handler<'a> {
...
+
+    // Stub for get_attestation() pseudo syscall
+    // See: https://github.com/enarx/enarx-keepldr/issues/31
+    pub fn get_att(&mut self) -> sallyport::Result {
+        self.trace("get_att", 0);
+
+        let rep = sallyport::Reply {
+            ret: [0.into(), SGX_TECH.into()],
+            err: Default::default(),
+        };
```
npmccallum, line 579: You want this instead:

```
let rep: sallyport::Reply = Ok([0.into(), SGX_TECH.into()]).into();
```

npmccallum (changes requested):

internal/shim-sgx/src/handler.rs
lkatalin: Thanks! I was looking for this and couldn't find the right way.

connorkuehl:
What do we gain from testing the stub?

lkatalin:
@connorkuehl Mostly the ability to know when I've messed something up as I fill in the implementation. : )

connorkuehl:
> @connorkuehl Mostly the ability to know when I've messed something up as I fill in the implementation. : )

I think that's only true if the test isn't also expected to change as the implementation is filled in:

```rust
   // TODO: Our current two success conditions are a technology value indicating SGX,
    // as get_att() is implemented for each technology.
    // or an error of ENOSYS. This will be updated as implementation is filled in.
    return !(size == -1 && errno == ENOSYS);
    return !((size == -1 && errno == ENOSYS) || (size == 0 && technology == SGX_TECH));
```

lkatalin:
Well, specifically I'm going to be filling in the SGX component next. So, if this test breaks, it'll be obvious that something is wrong with the SGX implementation of `get_att()` from the shim to the host. I can't currently add the other success conditions for the test because they won't be true yet, ex. SEV does not yet return the expected value that we'll eventually test for.

connorkuehl:
> Well, specifically I'm going to be filling in the SGX component next. So, if this test breaks, it'll be obvious that something is wrong with the SGX implementation of `get_att()` from the shim to the host. I can't currently add the other success conditions for the test because they won't be true yet, ex. SEV does not yet return the expected value that we'll eventually test for.

From my reading, it looks like as soon as you implement `get_att` for SGX, this test will break immediately anyways because a correct implementation of `get_att` will return a non-zero value that corresponds to the bytes filled into `buf`.

npmccallum:
This test has to run under four backends:
  * `sev`: returns success with the SEV attestation
  * `sgx`: returns success with the SGX attestation
  * `kvm`: returns `ENOSYS`
  * `nil`: returns `ENOSYS`

lkatalin:
@connorkuehl You're right, I was going to update that. I could do `(size >= 0 && technology == SGX_TECH)`. The point right now is just to test that `technology` returns with `2` for SGX.

npmccallum:
> @connorkuehl You're right, I was going to update that. I could do `(size >= 0 && technology == SGX_TECH)`. The point right now is just to test that `technology` returns with `2` for SGX.

I would prefer explicit branches for each condition. Otherwise the logic will get hairy quickly.

lkatalin:
I fixed this test failure by stubbing in the expected values in the SEV case.

internal/shim-sev/src/syscall.rs
```diff
@@ -196,10 +199,10 @@ pub fn syscall_rust(
         // Enarx syscalls
 
         // Fixme
-        let ret: sallyport::Result = Err(libc::ENOSYS);
+        let ret: sallyport::Result = Ok([0.into(), SEV_TECH.into()]);
 
         match ret {
-            Err(e) => (e.checked_neg().unwrap() as usize, orig_rdx),
```
haraldh, line 202: If there is only one return value it must return orig_rdx in the second.

haraldh (changes requested):

internal/shim-sev/src/syscall.rs
lkatalin: Got it, I put it back the way it was.

haraldh (approved):

tests/bin/get_att.c
```diff
@@ -9,7 +13,15 @@ int main(void) {
...
+        return !(size == -1 && errno == ENOSYS);
+
+    else if (technology == SEV_TECH)
+        return !(size >= 0);
+
+    else if (technology ==  SGX_TECH)
+        return !(size >= 0);
+
+    else
+        return 1;
```
npmccallum, line 26: I think we want to invert these branches. Match on the `size` value first. Then `errno` or `technology`. You `MUST NOT` inspect `technology` if an error occurred.

npmccallum (changes requested):

tests/bin/get_att.c
lkatalin: That is very reasonable.

internal/shim-sev/src/syscall.rs
```diff
@@ -196,7 +199,7 @@ pub fn syscall_rust(
         // Enarx syscalls
 
         // Fixme
-        let ret: sallyport::Result = Err(libc::ENOSYS);
+        let ret: sallyport::Result = Ok([0.into(), SEV_TECH.into()]);
```
connorkuehl, line 202:
```suggestion
        let ret: sallyport::Result = match nr {
            0xEA01 => Ok([0.into(), SEV_TECH.into()]),
            _ => Err(libc::ENOSYS),
        };
```

SGX has a `SYS_GETATT` but I don't see one defined for SEV.

tests/bin/get_att.c
```diff
@@ -9,7 +15,12 @@ int main(void) {
 
     ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
 
-    // TODO: Currently this should return ENOSYS. However, this will change
-    // as get_att() is implemented for each technology.
-    return !(size == -1 && errno == ENOSYS);
+    if (size >= 0)
+	return !(NO_KEEP <= technology <= SEV_TECH);
```
connorkuehl, line 19:
```suggestion
    if (size >= 0) {
	switch (technology) {
	case NO_KEEP:
	case SEV:
	case SGX: 
	    /* fallthrough is intentional */
	    return 0;
	default: return 1;
	}
    }
```

I think something like the above is easier to follow because it requires less mental gymnastics for operator precedence.

connorkuehl (changes requested):

tests/bin/get_att.c
connorkuehl: Also I think the boolean logic the way it is right now will always unconditionally return 0

```
return !(NO_KEEP <= technology <= SEV_TECH)
return !((0 <= 0..INTMAX) <= 0) 
return !(1 <= 1)
return !(1)
return 0
```

`technology` being unsigned means it will never be less than `NO_KEEP`

lkatalin: Ooh, good catch. I had forgotten `technology` is unsigned.

connorkuehl (dismissed):

haraldh (dismissed):

npmccallum (approved):

connorkuehl (approved):

PR #121, opened 2020-10-28
Implement close() syscall for SEV shim.

### LINKED ISSUE #120: Implement close() syscall for SEV shim.

greyspectrum:
Resolves #120

### DIFF
--- a/internal/shim-sev/src/syscall.rs
+++ b/internal/shim-sev/src/syscall.rs
@@ -219,6 +219,12 @@ pub fn exit(status: libc::c_int) -> ! {
 }
 
 /// syscall
+pub fn close(status: libc::c_int) -> ! {
+    eprintln!("SC> close({})", status);
+    hostcall::shim_exit(status);
+}
+
+/// syscall
 pub fn exit_group(status: libc::c_int) -> ! {
     eprintln!("SC> exit_group({})", status);
     hostcall::shim_exit(status);


internal/shim-sev/src/syscall.rs
```diff
@@ -218,6 +218,12 @@ pub fn exit(status: libc::c_int) -> ! {
     hostcall::shim_exit(status);
 }
 
+/// syscall
+pub fn close(status: libc::c_int) -> ! {
+    eprintln!("SC> close({})", status);
+    hostcall::shim_exit(status);
```
haraldh, line 224: `close` != `exit`

haraldh (changes requested):

connorkuehl (changes requested):
Thanks for the PR, @greyspectrum!

In Enarx, we can broadly group system calls into two groups:

* Unproxied syscalls
* Proxied syscalls

You may already know, but since our shims serve as the kernel inside of a TEE, sometimes it's unable to fulfil a request; in this case, the shim will proxy the call to the host.

Right now, neither shim has support for handling files independently of the host, so syscalls related to files, such as `close()` will need to be proxied to the host.

For your feature, you'll be implementing the `close` syscall as a proxied syscall.

The entry point for your syscall is in the `syscall` module (src/syscall.rs).

My suggestions are:

1. Add an entry for your `close()` syscall to the syscall dispatch table in the `syscall_rust` function. ([example](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sev/src/syscall.rs#L151))
2. `man 2 close` suggests that the number of arguments you have for your syscall are correct; however, I'd suggest changing the name from `status` to `fd` because the syscall argument is a file descriptor from userspace.
3. In your `close()` implementation, you'll need to acquire the `HOST_CALL` lock because in order to proxy the system call, the shim must write a `sallyport::Request` to a region of memory that it shares with the host. So it's important to be thread-safe and acquire the lock. ([example](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sev/src/syscall.rs#L291))
4. Now you can get access to the shared memory region by calling the `as_mut_block()` method on the `HOST_CALL` lock guard. ([example](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sev/src/syscall.rs#L292))
5. The `close()` syscall only needs to provide the fd, so you can set up the proxied syscall with something as simple as: `block.msg.req = request!(libc::SYS_close => fd);` ([example](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sev/src/syscall.rs#L299))
6. Finally, proxy the syscall ([example](https://github.com/enarx/enarx-keepldr/blob/master/internal/shim-sev/src/syscall.rs#L301))

One last suggestion would be to include your `close()` test in this PR after you've implemented the syscall. This can be a separate commit.

If you have any questions, please don't hesitate to drop by our chat

lkatalin:
@greyspectrum This has been handled in #131 , so can be closed. But the good news is that the `close()` test should hopefully now pass after a rebase.

PR #122, opened 2020-10-28
rework of shim-sev: handle multiple return values for syscalls

haraldh:
Revert the previous patch and introduce a much simpler one.

shim-sev: handle multiple return values for syscalls

`rdx` is sometimes used as a secondary return value.

This patch lets `syscall_rust` return a double u64 struct instead of u64,
which will use `rax` for the first 64 bits and `rdx` for the last 64 bit.

Because `rdx` is not marked as `clobber` for the musl syscall macro,
the original value has to be returned, if `rdx` is unused as a return
value.

This patch also adds a dummy path for the enarx syscalls, which can make
use of the second return value.

### DIFF
--- a/internal/shim-sev/src/syscall.rs
+++ b/internal/shim-sev/src/syscall.rs
@@ -29,15 +29,6 @@ const FAKE_GID: usize = 1000;
 #[inline(never)]
 #[naked]
 pub unsafe fn _syscall_enter() -> ! {
-    let rdi: usize;
-    let rsi: usize;
-    let rdx: usize;
-    let r10: usize;
-    let r8: usize;
-    let r9: usize;
-    let nr: usize;
-    let rbx: usize;
-
     use crate::gdt::{USER_CODE_SEGMENT, USER_DATA_SEGMENT};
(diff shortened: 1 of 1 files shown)


haraldh:
Best to review commit by commit.

lkatalin (approved):
Seems like a good solution to me.

npmccallum (commented):
@haraldh It looks like you're inventing new types (`X8664DoubleReturn`), but we already have similar types in `sallyport`. Why not use those?

haraldh:
> @haraldh It looks like you're inventing new types (`X8664DoubleReturn`), but we already have similar types in `sallyport`. Why not use those?

@npmccallum I don't see any FFI-safe type with two u64/usize struct members in `sallyport`

npmccallum:
> @npmccallum I don't see any FFI-safe type with two u64/usize struct members in `sallyport`

https://github.com/enarx/enarx-keepldr/blob/master/src/sallyport/mod.rs#L92

npmccallum (approved):

Issue #31, opened 2020-10-01
Design an attestation syscall


haraldh:
# Attestation Syscall - User Space

syscall number `0xEA01`

## NAME

get_attestation - get the attestation from the CPU

## SYNOPSIS

For `C`
```C
ssize_t get_attestation(void *nonce, size_t nonce_length, void *buf, size_t buf_length, size_t *technology);
```

Note: a `C` wrapper has to store the second return value in `rdx` into `technology` and a negative `rax` in errno.

## RETURN VALUE

On error a negative number is returned as `-errno`.

On success, this call returns the number of bytes placed in buf.

`technology` is filled in with
| value | TEE technology identifier  |
|--------|---------------|
| 0       | NONE  |
| 1       | SEV  |
| 2       | SGX |

### Inputs

1. `rdi` - input data buffer (const)
2. `rsi` - input data buffer length
3. `rdx` - output data buffer
4. `r10` - output data buffer length

### Outputs

1. `rax` - number of bytes written or negative errno (per Linux convention)
2. `rdx` - TEE technology identifier (enum)

### Special Case

When `rdx` (input) is `0` (i.e. `NULL`):
* `r10` is ignored
* no data is written (since the buffer is non-existent)
* other outputs are the same as if the call had succeeded with a large enough buffer

### Format of `buf`

Serialization is done on the host and deserialization is done in the Wasmtime code payload. Both have access to full `std`, so
all features of `serde` should be used to ensure a safe data transfer from the untrusted host data.

`rdx` or `technology` from the`getatt()` call decides which of the union members to use.

Example union for `buf`:
```rust
#[repr(C)]
pub union Attestation {
    Sev_PrivateKeyInfo,
    Sgx_Attestation,
    // … to be extended
}
```

Suggested serde formats:
* json - best maintained - most languages have bindings
* cbor - smaller transfer - standardized, a lot of languages have bindings
* bincode - smaller transfer

## ERRORS

EFAULT: buf extends outside the process's allocated address space.

EINVAL: bufsiz is not positive.

EIO: an error occurred getting the attestation

EMSGSIZE: provided buffer too small

# Attestation Syscall - Shim to Host

Because the method of attestation differs for every architecture, the internal syscall handling differs for each.

## SEV - pre attestation - Shim

`buf` provided to the host is the shim physical address, where the host places encrypted bytes received from the tenant.
This is not an address in the unencrypted `sallyport::Block`.

[...]

PR #133, opened 2020-11-05
Stub in shim --> host implementation of get_att() for SGX

### LINKED ISSUE #91: Implement stub for sgxatt() pseudo-syscall in SGX (shim => host)
The `sgxatt()` pseudo-syscall is the SGX-specific version of `getatt()` with the same syscall number (`0xEA01`) from the shim to the host layer. It should at first return a synthetic Quote of the proper length.

lkatalin:
Closes #91 

This PR:
- Updates the function signature for `get_attestation()` in the internal `syscall` mod, the sev handler, and the sgx handler to pass the four parameters mentioned in #31 .
- Adds logic in the SGX shim handler to make two calls out to the host before returning a Quote to the code layer: one to request a TargetInfo from the host, and the second to request a Quote from the host. The first is a necessary prerequisite for the second.
- Adds an attestation mod in the sgx backend on the host side to handle proxied `get_att()` syscalls from the sgx shim. This currently returns synthetic values, but will communicate with the AESM daemon to get real TargetInfo and Quote values in the next PR.
- Adds logic in the sgx backend mod to route `get_att()` syscalls to the above attestation mod.
- Adds a test to request a Quote and check for expected values in the returned synthetic Quote.

### DIFF
--- a/internal/shim-sgx/src/handler.rs
+++ b/internal/shim-sgx/src/handler.rs
@@ -7,8 +7,10 @@ use core::fmt::Write;
 use sallyport::{request, Block, Cursor, Request};
 use sgx::types::ssa::StateSaveArea;
 use sgx_heap::Heap;
-use syscall::{SyscallHandler, ARCH_GET_FS, ARCH_GET_GS, ARCH_SET_FS, ARCH_SET_GS, SGX_TECH};
-use untrusted::{AddressValidator, UntrustedRef, ValidateSlice};
+use syscall::{
+    SyscallHandler, ARCH_GET_FS, ARCH_GET_GS, ARCH_SET_FS, ARCH_SET_GS, SGX_TECH, SYS_GETATT,
+};
+use untrusted::{AddressValidator, UntrustedRef, UntrustedRefMut, ValidateSlice};
 
 pub const TRACE: bool = false;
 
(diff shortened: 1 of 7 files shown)


src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,87 @@
...
+/// for the QE, or a Quote verifying a Report.
+pub fn get_attestation(
+    nonce: usize,
+    _nonce_len: usize,
+    buf: usize,
+    _buf_len: usize,
+) -> Result<usize, Error> {
+    // If the output buffer length is 0, returns the size required for the output
+    // buffer to verify a Report.
+    if let 0 = buf {
```
haraldh, line 22: why not `if 0 == buf` or `if buf.eq(0)` ???

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,87 @@
...
+        return Ok(QUOTE_SIZE);
+    }
+
+    // If the nonce is 0 (NULL), fills the Target Info of the QE into the output buffer
+    // and returns the number of bytes written.
+    if let 0 = nonce {
```
haraldh, line 28: same as above

src/backend/sgx/mod.rs
```diff
@@ -198,6 +203,21 @@ impl super::Thread for Thread {
...
+                        let result = unsafe {
+                            get_attestation(
+                                self.block.msg.req.arg[0].into(),
+                                self.block.msg.req.arg[1].into(),
+                                self.block.msg.req.arg[2].into(),
+                                self.block.msg.req.arg[3].into(),
+                            )?
+                        };
+
+                        // TODO: Probably a better rdx to return here
```
haraldh, line 216: doesn't matter

lkatalin: @haraldh I can remove the comment - is there anything else to change in the PR?

haraldh: not besides `internal/shim-sgx/src/handler.rs` ... and btw, you should change the signature of the `get_attestation` method in the trait also to match the arguments used in the syscall definition (like all the other syscalls). Then you don't have to use `self.aex.gpr.*`

lkatalin: @haraldh I updated the function signatures for `get_attestation()` - is this all you meant by "not besides `handler.rs`"?

internal/shim-sgx/src/handler.rs
```diff
@@ -367,8 +369,19 @@ impl<'a> SyscallHandler for Handler<'a> {
...
+        &mut self,
+        nonce: *const libc::c_void,
+        nonce_len: libc::size_t,
+        buf: *const libc::c_void,
+        buf_len: libc::size_t,
+    ) -> sallyport::Result {
         self.trace("get_att", 0);
-        Ok([0.into(), SGX_TECH.into()])
+
+        let req = request!(SYS_GETATT => nonce, nonce_len, buf, buf_len);
```
haraldh, line 381: you are still missing the Block Cursor action before and after this

haraldh (changes requested):

internal/shim-sgx/src/handler.rs
lkatalin: Ah okay, I figured since it is the stub I would do it in the next PR, but I might as well do it here.

lkatalin: @haraldh I updated it to use the Block. There may be a cleaner way to do it, but it should work... it copies the `nonce` to the block first, then allocates space for the `buf` so that it can get a pointer to the area right after the nonce. Afterwards it didn't seem like I can use the same pointers, so I made a new block, allocated the `nonce` to skip that area and get a pointer to the `buf` to copy data out of the block.

lkatalin: Actually, I am thinking maybe I should test it before this is merged.

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    // If the nonce is 0 (NULL), fills the Target Info of the QE into the output buffer
+    // and returns the number of bytes written.
+    if nonce == 0 {
+        // TODO: Populate with real Target Info (https://github.com/enarx/enarx-keepldr/issues/92).
+        let b: &mut [u8] = unsafe { from_raw_parts_mut(buf as *mut u8, _buf_len) };
+
+        // NOTE: While this doesn't return the number of bytes written, it will error if the two
+        // slices are of differing size.
+        b.copy_from_slice(&TMP_TI);
```
haraldh, line 34: well, it panics, which is maybe not that nice... returning an error is nicer :)

src/backend/sgx/attestation.rs
```diff
@@ -0,0 +1,87 @@
...
+const TMP_QUOTE: [u8; QUOTE_SIZE] = [0u8; QUOTE_SIZE];
+
+/// Returns either the size required for the output buffer, or the number of bytes written to the
+/// output buffer. Depending on parameters, the output buffer will be filled with the Target Info
+/// for the QE, or a Quote verifying a Report.
+pub fn get_attestation(
+    nonce: usize,
+    _nonce_len: usize,
+    buf: usize,
+    _buf_len: usize,
```
haraldh, line 18: No need for an `_`, if you are using the length.

haraldh (changes requested):

lkatalin:
Okay, this version works for reading data to and from the shim. :+1:

lkatalin:
Rewrote the `handler` logic to deal with validating NULL pointers and added two tests.

internal/shim-sgx/src/handler.rs
```diff
@@ -367,8 +369,53 @@ impl<'a> SyscallHandler for Handler<'a> {
...
+        let c = self.new_cursor();
+        let (c, shim_nonce_ptr) = c.copy_from_slice(&tmp_report).or(Err(libc::EMSGSIZE))?;
+        let (_, shim_buf_ptr) = c.alloc::<u8>(buf_len).or(Err(libc::EMSGSIZE))?;
+        let req = request!(SYS_GETATT => shim_nonce_ptr.as_ptr(), tmp_report.len(), shim_buf_ptr.as_ptr(), buf_len);
+        let result = unsafe { self.proxy(req)? };
+
+        // Pass Quote back to code layer in buf
+        let c = self.new_cursor();
+        let (c, _) = c.alloc::<u8>(tmp_report.len()).or(Err(libc::EMSGSIZE))?;
+        c.copy_into_slice(buf_len, buf.as_mut(), buf_len)
```
haraldh, line 415:
```suggestion
        let result_len: usize = ret[0].into();
        
        if result_len > buf_len {
            self.attacked()
        }

        c.copy_into_slice(buf_len, buf.as_mut(), result_len)
```

haraldh (changes requested):

internal/shim-sgx/src/handler.rs
lkatalin: Good idea.

src/backend/sgx/mod.rs
```diff
@@ -20,11 +21,15 @@ use std::convert::TryInto;
...
 use std::sync::{Arc, RwLock};
 
+mod attestation;
 mod data;
 mod shim;
 
 const SHIM: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/bin/shim-sgx"));
 
+// See https://github.com/enarx/enarx-keepldr/issues/31
+const SYS_GETATT: usize = 0xEA01;
```
haraldh, line 31: Note: In a followup patch I will collect all these syscall consts and put them in a central place.

haraldh (approved):

connorkuehl (approved):

PR #139, opened 2020-11-10
Allow all non-copyleft licenses

mbestavros:


haraldh:
Why would we allow non-OSI/FSF licenses?

haraldh (changes requested):
See https://github.com/enarx/enarx-keepmgr/pull/13#pullrequestreview-526971216

PR #140, opened 2020-11-10
Generate a Quote for SGX attestation using AESM daemon

### LINKED ISSUE #92: Implement logic for sgxatt() pseudo-syscall in SGX
This syscall will:

- [x] Call out to the host to retrieve a QE TargetInfo
- [x] Call EREPORT with the TargetInfo and a nonce as parameters to obtain the enclave's Report
- [x] Call out to the host to retrieve a Quote based on the Report

Also: 
- [x] Logic should be included to build a suitable .proto file

- When called with NULL input, the host should interpret this as an `initQuoteRequest` to the AESM daemon and should pass the resulting QE TargetInfo back to the shim. 
- When called with a Report as input, the host should interpret this as a `getQuoteRequest` to the AESM daemon and should pass the resulting Quote back to the shim.

lkatalin:
In this PR:

- [x] Pass the Report back to host to generate a Quote
- [x] send Quote to code layer

~~This will be rebased onto #163 once that is merged. Ignore the first two commits for this reason, as they are a separate PR.~~
- [x] This will be rebased onto #179  once that is merged.

The `get_quote()` values are hardcoded and need to be generated in a more straightforward way, though this will require research, possibly in a follow-up debt issue.

Closes #92 

Depends on: #133 , https://github.com/enarx/sgx/pull/7 , #163 , #179


internal/shim-sgx/src/handler.rs
```diff
@@ -393,10 +432,26 @@ impl<'a> SyscallHandler for Handler<'a> {
...
+                        let f = read_u64(&ti[32..40], ByteOrder::LittleEndian).unwrap();
+                        let x = read_u64(&ti[40..48], ByteOrder::LittleEndian).unwrap();
+                        let flags = Flags::from_bits(f).unwrap();
+                        let xfrm = Xfrm::from_bits(x).unwrap();
+                        let att = Attributes::new(flags, xfrm);
+                        target_info.mrenclave.copy_from_slice(&ti[0..32]);
+                        target_info.attributes = att;
+
+                        let data = ReportData([0u8; 64]);
+                        let _report: Report = unsafe { target_info.get_report(&data) };
```
haraldh, line 452: I thought we agreed on not parsing the get_attestation data in the shims, but only pass them through as byte buffers?

lkatalin: Hmm, maybe that should be revised. The Report has to be generated in the shim, because it can only be generated inside an enclave.

haraldh: ah, oh, ok

@npmccallum your thoughts?

tests/bin/sgx_get_att_quote.c
```diff
@@ -5,14 +5,11 @@
  * Quote from get_attestation() and check that the first bytes of
  * the returned Quote in buf match expected values. */
 
-// TODO: Update these to match real Quote values.
-// See https://github.com/enarx/enarx-keepldr/issues/92.
-const unsigned char expected[512] = { 44, 44, 44 };
+// TODO: Determine expected Quote value
```
haraldh, line 8: I guess this value would change every time we change the shim-sgx, right?

haraldh (dismissed):

tests/bin/sgx_get_att_quote.c
lkatalin: Yes, it would. So I don't think it would make sense to check it in its entirety, at least until the shim is very stable. However, there may be some values that are always the same (I think some QE data may be embedded in each Quote). I plan to review that and see if maybe we can just check those values, if they exist.

haraldh:
This needs a rebase for the tests to succeed, I guess.

haraldh (dismissed):

internal/shim-sgx/src/handler.rs
```diff
@@ -447,20 +450,23 @@ impl<'a> SyscallHandler for Handler<'a> {
...
-        let _report: Report = unsafe { target_info.get_report(&data) };
+        let report: Report = unsafe { target_info.get_report(&data) };
 
         // Request Quote from host
-        // TODO: Send a real Report. See https://github.com/enarx/enarx-keepldr/issues/92.
-        let tmp_report = [0u8; 512];
+        let report_len = core::mem::size_of::<Report>();
+        let report_slice = &[report];
+        let (_, report_bytes, _) = unsafe { report_slice.align_to::<u8>() };
+        assert_eq!(report_bytes.len(), report_len);
```
haraldh, line 459: The `assert_eq` should probably be a `debug_assert_eq`, because we can assume the behavior doesn't change in   release version.

I give you, this is a lot less code to write and read, than:
```rust
        let report_bytes = unsafe {
            core::slice::from_raw_parts(
                report_slice.as_ptr() as *const _ as *const u8,
                core::mem::size_of::<Report>(),
            )
        };
```
although I first wondered why the align to `u8` ... but yeah, shorter to write :+1:

haraldh (approved):

src/backend/sgx/attestation.rs
```diff
@@ -81,15 +82,80 @@ fn get_ti(out_buf: &mut [u8]) -> Result<usize, Error> {
...
     );
-    out_buf.copy_from_slice(&SGX_DUMMY_QUOTE);
-    SGX_QUOTE_SIZE
+
+    // If unable to connect to the AESM daemon, return dummy value
+    let mut stream = match UnixStream::connect(AESM_SOCKET) {
+        Ok(s) => s,
+        Err(_) => {
+            out_buf.copy_from_slice(&SGX_DUMMY_QUOTE);
+            return Ok(SGX_QUOTE_SIZE);
```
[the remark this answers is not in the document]
lkatalin: Good suggestion. I've made this follow-up issue here: #228 .

mbestavros (approved):

haraldh (approved):

Issue #135, opened 2020-11-06
Update Sallyport syscall test to be more widely applicable


lkatalin:
This unit test currently expects certain fd's to not be in use, as it hard codes expected fd values. Consequently, the test fails in the SGX case when a connection is made to the Unix socket for the AESM daemon. A temporary fix is to not run this test for SGX, but there must be better ways.

PR #147, opened 2020-11-11
fix(sallyport): Don't hardcode dup'ed file descriptor value

haraldh:
If additional filedescriptors are open, the `syscall` test fails,
because it assumes, that the return value of dup() is 3.

This patch removes the hardcoded value and checks,
if the return value is greater 0 and then closes the dup'ed fd.

Fixes: https://github.com/enarx/enarx-keepldr/issues/135


lkatalin (approved):
Thanks!

connorkuehl (approved):

PR #148, opened 2020-11-11
feat(tests): Better output on failure

haraldh:
*  feat(tests): Better output on failure
    
    The integration tests output very long `assert_eq` comparisons, if the
    output doesn't match the expected. To reduce the clutter, only use
    `assert_eq` if the size is smaller than 100 first, then assert on the
    length and finally do the long `assert_eq`.
    
    To indicate what assertion failed, add descriptive messages to the
    `assert_eq` macros.
    
    Also, in the case of no expected stderr, print out any stderr received,
    to help debug any failures.


lkatalin (approved):
LGTM, thanks.

connorkuehl (changes requested):
I think we should throw `100` into a const and refer to the const variable

haraldh:
Revised the whole patch again for more readability.

connorkuehl (approved):

PR #151, opened 2020-11-11
fix(sallyport): Remove obsolete doc section

haraldh:
This doc section seems to be a leftover, when some module was called `proxy`.


lkatalin (approved):

connorkuehl (approved):
Indeed. 

`proxy` was the name of a module where the sallyport stuff lived but it was essentially the entirety of the `sallyport` crate and was the only module, so its contents were simply pulled out to be top-level code in the `sallyport` crate.

So spake a budding Enarx loremaster.

And he approves thusly

PR #152, opened 2020-11-11
fix(syscall): Use correct pointer for clock_gettime

haraldh:
While the address of the whole array should be the same
as the first element, don't make such assumptions and
use the pointer of the first element.


lkatalin:
I'm curious how they could be different. When does this happen?

haraldh:
> I'm curious how they could be different. When does this happen?

When the `.as_ptr()` method is implemented differently, or the types returned are different to the ones expected.

lkatalin (approved):

ambaxter (approved):
LGTM. It broke my brain for a second, but I understand what went wrong now.

Issue #23, opened 2020-09-15
Fix the sallyport block size


connorkuehl:
As a stopgap measure for hosting the wasmldr, we ended up increasing the size of the sallyport block by a large amount. This change must be effectively reverted immediately and implemented properly.

This PR merged the change: https://github.com/enarx/enarx-keepldr/pull/22

haraldh:
@connorkuehl  @npmccallum why not make short reads instead? The rust implementation for `read_all()` just reads until `0` or an error occurred.

Alternatively we can do the same in our proxy `read()` in the shim and fill the buffer, until `0` or error.

Issue #156, opened 2020-11-11
shim-sev: use multiple blocks for host <-> shim communication


haraldh:
When multi-threading is implemented multiple buffers need to be
available for concurrent syscalls.

PR #157, opened 2020-11-11
Fix the sallyport blocksize

haraldh:
*   fix(syscall): Read/Write in chunks of Block::buf_capacity
    
    Read and Write from the host in chunks of `Block::buf_capacity()`,
    if the requested length is greater.
    While this might cause problems with blocking file descriptors,
    it is very unlikely though.
   
*   fix(sallyport): Fix the sallyport block size
    
    The sallyport block size was set 512 * 4096 bytes, whereas
    the space was never allocated in the kvm backend.
    
    This patch sets the capacity to hold at least a maximum UDP packet.
    
    When multi-threading is implemented multiple buffers need to be
    available for concurrent syscalls.
    Related: https://github.com/enarx/enarx-keepldr/issues/156
    
    Fixes: https://github.com/enarx/enarx-keepldr/issues/23

*   feat(tests): Add read_udp test
    
    This test tries to read 65507 in one go.
    It fails, if there was a short read.
    
    While this test does not explicitly tests, if the shim has
    proxied this to the host with length 65507, an strace shows
    that it has.
    
    A future PR will change this to use real UDP.
    Related: https://github.com/enarx/enarx-keepldr/issues/154
    
    To test, if bigger reads can be still processed, increase
    the `echo` test size to 2MB.

### DIFF
--- a/internal/syscall/src/lib.rs
+++ b/internal/syscall/src/lib.rs
@@ -9,7 +9,7 @@
 use core::convert::TryInto;
 use core::mem::MaybeUninit;
 use primordial::Register;
-use sallyport::{request, Cursor, Request, Result};
+use sallyport::{request, Block, Cursor, Request, Result};
 use untrusted::{AddressValidator, UntrustedRef, UntrustedRefMut, Validate, ValidateSlice};
 
 /// `get_attestation` syscall number
@@ -192,6 +192,9 @@ pub trait SyscallHandler: AddressValidator + Sized {
 
         let c = self.new_cursor();
 
+        // Limit the read to `Block::buf_capacity()`
+        let count = usize::min(count, Block::buf_capacity());
+
         let (_, hostbuf) = c.alloc::<u8>(count).or(Err(libc::EMSGSIZE))?;
         let hostbuf = hostbuf.as_ptr();
         let host_virt = self.translate_shim_to_host_addr(hostbuf);
@@ -205,7 +208,7 @@ pub trait SyscallHandler: AddressValidator + Sized {
         }
 
         let c = self.new_cursor();
-        c.copy_into_slice(count, buf.as_mut(), result_len)
+        c.copy_into_slice(count, buf[..count].as_mut(), result_len)
             .or(Err(libc::EFAULT))?;
 
         Ok(ret)
@@ -234,6 +237,9 @@ pub trait SyscallHandler: AddressValidator + Sized {
 
     /// syscall
     fn write(&mut self, fd: libc::c_int, buf: UntrustedRef<u8>, count: libc::size_t) -> Result {
+        // Limit the write to `Block::buf_capacity()`
+        let count = usize::min(count, Block::buf_capacity());
+
         let buf = buf.validate_slice(count, self).ok_or(libc::EFAULT)?;
 
         let c = self.new_cursor();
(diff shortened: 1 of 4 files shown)


internal/syscall/src/lib.rs
```diff
@@ -235,13 +238,41 @@ pub trait SyscallHandler: AddressValidator + Sized {
...
-        let (_, buf) = c.copy_from_slice(buf.as_ref()).or(Err(libc::EMSGSIZE))?;
-        let buf = buf.as_ptr();
-        let host_virt = self.translate_shim_to_host_addr(buf);
+        let mut to_write = count;
+        let mut written = 0;
 
-        let ret = unsafe { self.proxy(request!(libc::SYS_write => fd, host_virt, count))? };
-        Ok(ret)
+        loop {
+            let count = usize::min(to_write, Block::buf_capacity());
```
lkatalin, line 245: Why do we need to specify this twice?

src/sallyport/mod.rs
```diff
@@ -5,6 +5,12 @@ use core::mem::MaybeUninit;
 use core::ptr::NonNull;
 use primordial::{Page, Register};
 
+/// The maximum size of a UDP packet
+///
+/// The maximum UDP message size is 65507, as determined by the following formula:
+/// 0xffff - (sizeof(minimal IP Header) + sizeof(UDP Header)) = 65535-(20+8) = 65507
```
lkatalin, line 11: Thanks for good documentation.

tests/bin/read_udp.c
```diff
@@ -0,0 +1,15 @@
...
+
+#include "libc.h"
+
+int main(void) {
+    char buf[65507];
+
+    ssize_t out = read(STDIN_FILENO, buf, sizeof(buf));
+
+    if (out != sizeof(buf))
+        return -1;
```
lkatalin, line 12: Are we not usually returning `1` on error in these tests? Not sure if this will matter for consistency.

lkatalin (commented):
Had two questions, otherwise looks good.

internal/syscall/src/lib.rs
connorkuehl: This shadows the `count` parameter which is immutable. If he hadn't shadowed it, this wouldn't compile unless he made the parameter mutable which is not desirable.

lkatalin: Ahh, sorry, I totally spaced out on the fact that the other `count` was in `read()` and not `write()`. Makes sense now!

tests/bin/read_udp.c
lkatalin: I take this back. Looks like we return a variety of non-zero values in the other tests.

lkatalin (approved):

internal/syscall/src/lib.rs
```diff
@@ -235,13 +238,41 @@ pub trait SyscallHandler: AddressValidator + Sized {
...
-        let c = self.new_cursor();
-        let (_, buf) = c.copy_from_slice(buf.as_ref()).or(Err(libc::EMSGSIZE))?;
-        let buf = buf.as_ptr();
-        let host_virt = self.translate_shim_to_host_addr(buf);
+        let mut to_write = count;
+        let mut written = 0;
 
-        let ret = unsafe { self.proxy(request!(libc::SYS_write => fd, host_virt, count))? };
-        Ok(ret)
+        loop {
```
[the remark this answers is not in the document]
haraldh: Yeah... you are right.

haraldh:
Rebased, updated `write()` and removed the timeout increase in the integration tests.

haraldh:
And also corrected `writev()`  for short writes.

connorkuehl (approved):

npmccallum (approved):

PR #165, opened 2020-11-12
feat(shim-sev): Use named parameters in asm! blocks

haraldh:
Named parameters make reading `asm!()` blocks a lot easier and are less
error prone to changes made in the arguments.


connorkuehl (approved):
Definitely!

npmccallum (approved):

PR #173, opened 2020-11-13
feat(syscall): Collect all syscall constants

haraldh:
The syscall constants are spread all over the place and named
inconsistently. This patch fixes that.


lkatalin (approved):
This is a good patch.

connorkuehl (approved):

PR #176, opened 2020-11-16
Fix spurious failure in sgx_get_att_quote test if not running on SGX

connorkuehl:
- Factor out TEE enum to its own header
- Return success if not running on SGX in sgx_get_att_quote test


ambaxter (approved):

lkatalin (dismissed):
Thanks!

lkatalin (approved):

haraldh (approved):

PR #188, opened 2020-11-23
fix build on nightly

haraldh:
*   fix(keepldr): suppress clippy warnings
    
    Suppress clippy warnings about `Result<()>`, where the functions never
    return an `Err()` for now.

*   fix(shim-sev): Function crt0setup does not need Result as return
    
    `crt0setup()` never returns an `Err()`, so remove `Result` from the
    return parameter, otherwise clippy will complain.

*   fix(build): Remove unnecessary return value
    
    `rerun_src()` doesn't need a return value, so remove it, because clippy
    will complain.
    
*   fix(shim-sev): remove `naked` attribute from usermode fn
    
    The `#[naked]` caused a rustc segfault. Remove it, because it was the
    wrong attribute anyway.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,14 +27,12 @@ fn find_files_with_extensions<'a>(
         .map(|x| x.path().to_owned())
 }
 
-fn rerun_src(path: impl AsRef<Path>) -> std::io::Result<()> {
+fn rerun_src(path: impl AsRef<Path>) {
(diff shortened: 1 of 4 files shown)


haraldh:
https://github.com/rust-lang/rust/issues/79350

ueno (approved):

mbestavros (approved):

PR #190, opened 2020-11-26
fix(sallyport): mark copy_into_slice unsafe

haraldh:
*   fix(sallyport): mark copy_into_slice unsafe
    
    `Cursor::copy_into_slice` has to be unsafe, otherwise it is possible to
    fill in a slice of e.g. bools with invalid values using safe rust code.
    
    This patch also adds a test calling `Cursor::copy_into_slice()` and
    simplifies the method by removing a parameter and calling existing
    methods.

### DIFF
--- a/src/sallyport/mod.rs
+++ b/src/sallyport/mod.rs
@@ -249,24 +249,23 @@ impl<'short, 'a: 'short> Cursor<'a> {
         Ok((c, dst))
     }
 
-    /// Copies data from a slice into the cursor buffer using self.alloc().
+    /// Copies data from a cursor buffer into a slice advancing the cursor.
+    ///
+    /// # Parameters
+    ///
+    /// * `src_len`: the amount of elements, the cursor is advanced after copying
+    /// * `dst`: the destination slice
+    ///
+    /// # Safety
+    /// The caller has to ensure the `Cursor` contains valid data of the type
+    /// of the destination slice.
     #[allow(dead_code)]
-    pub fn copy_into_slice<T: Copy>(
+    pub unsafe fn copy_into_slice<T: 'a + Copy>(
         self,
         src_len: usize,
         dst: &mut [T],
-        dst_len: usize,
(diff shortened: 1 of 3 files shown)


haraldh:
Needs https://github.com/enarx/enarx-keepldr/pull/188 first

connorkuehl (dismissed):

src/sallyport/mod.rs
```diff
@@ -250,23 +250,19 @@ impl<'short, 'a: 'short> Cursor<'a> {
     }
 
     /// Copies data from a slice into the cursor buffer using self.alloc().
```
lkatalin, line 252: Unrelated to this PR, but should this not be "copies from a cursor buffer into a slice"?

src/sallyport/mod.rs
```diff
@@ -566,4 +562,42 @@ mod tests {
...
+
+        assert_eq!(slab1, &[1, 2]);
+        assert_eq!(slab2, &[3, 4]);
+        assert_eq!(slab3, &[5, 6]);
+
+        let c = block.cursor();
+
+        let mut slab_all = [0usize; 3];
+
+        let c = unsafe { c.copy_into_slice::<usize>(4, &mut slab_all, 3) }?;
```
lkatalin, line 591: Why is the `src_len` a `4` here? That is shorter than the total slab 1-3 values but longer than `slab_all`.

src/sallyport/mod.rs
haraldh: :+1:

src/sallyport/mod.rs
haraldh: That is the amount of elements the cursor should advance after copying.

haraldh:
Ok, I didn't see the forest because of all the trees... simplified the method by removing `dst_len` and documenting the arguments.

lkatalin (approved):

connorkuehl (approved):

PR #192, opened 2020-11-26
Sev concurrent syscalls

haraldh:
*   feat(shim-sev): Use multiple sallyport::Block
    
    Because the setup area is limited in memory size, not each thread can
    get its own `sallyport::Block` to communicate with the host.
    
    Therefore a pool of blocks is reserved, which can be dynamically used
    by the shim indicating the actual block via the value passed in the
    IOPort.
    
    A future extension could reserve additional blocks via a enarx specific
    syscall.
    
    This patch also removes the now obsolete `backend::kvm::vm::Builder::Config`.

    Fixes: https://github.com/enarx/enarx-keepldr/issues/156
    
*   fix(shim-sev): rename ALIGN_ABOVE_2MB
    
    Rename `ALIGN_ABOVE_2MB` to `MAX_SETUP_SIZE` to reflect its meaning.

### DIFF
--- a/src/backend/kvm/vm/builder.rs
+++ b/src/backend/kvm/vm/builder.rs
@@ -2,8 +2,8 @@
 
 use super::*;
 use crate::backend::kvm::shim::BootInfo;
-
 use crate::binary::Component;
+use crate::sallyport::Block;
 
 use anyhow::Result;
 use kvm_ioctls::{Kvm, VmFd};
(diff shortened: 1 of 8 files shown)


haraldh:
Needs https://github.com/enarx/enarx-keepldr/pull/188 first

src/backend/kvm/vm/builder.rs
```diff
@@ -14,22 +14,45 @@ use primordial::Page;
...
+#[repr(C, align(4096))]
+pub struct SetupRegion {
+    pre: SetupRegionPre,
+    syscall_blocks: [Block; N_SYSCALL_BLOCKS],
 }
 
+pub const N_SYSCALL_BLOCKS: usize =
+    (MAX_SETUP_SIZE - size_of::<SetupRegionPre>()) / size_of::<Block>();
+
+pub struct Config;
```
connorkuehl, line 51: I think `Config` can be removed. Unless we want the number of syscall blocks to be configurable.

connorkuehl (changes requested):

src/backend/kvm/vm/builder.rs
haraldh: :+1:

haraldh:
also removed `backend::kvm::vm::Builder::Config`

connorkuehl (approved):

lkatalin (approved):

PR #170, opened 2020-11-13
feat(shim-sev): Implement get_attestation for SEV

haraldh:
Inject the secret and provide it with the get_attestation call.

Fixes: https://github.com/enarx/enarx-keepldr/issues/159

PR #194, opened 2020-11-26
feat(keepldr): Measure the contents of the keep for SEV

haraldh:
*   feat(keepldr): Measure the contents of the keep for SEV
    
    This patch adds a new command `measure`, which produces a measurement on
    the deployed keep and outputs relevant data as a json record,
    which can be used by clients to validate a keep.
    
    This patch implements the measurement for the SEV backend.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -115,11 +115,19 @@ struct Exec {
     code: PathBuf,
 }
 
+/// Get a report from a keep
+#[derive(StructOpt)]
+struct Report {
+    /// The payload to run inside the keep
+    code: PathBuf,
+}
+
 #[derive(StructOpt)]
 #[structopt(version=VERSION, author=AUTHORS.split(";").nth(0).unwrap())]
 enum Options {
     Info(Info),
     Exec(Exec),
+    Report(Report),
 }
 
 #[allow(clippy::unnecessary_wraps)]
@@ -136,6 +144,7 @@ fn main() -> Result<()> {
     match Options::from_args() {
         Options::Info(_) => info(backends),
         Options::Exec(e) => exec(backends, e),
+        Options::Report(e) => measure(backends, e),
     }
 }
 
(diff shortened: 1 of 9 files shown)


haraldh:
Needs https://github.com/enarx/enarx-keepldr/pull/170 first

connorkuehl:
I have been thinking about something like this too. As of now, our measured initial start is generated at runtime but is a static thing. We don't tune it in any way. We are, I think, wasting our time with generating the image at launch.

I wonder if it would be beneficial to create a tool that can generate a minimal Enarx image with our shim and wasmldr already loaded. Then we can simply measure the image and this measurement can be published, though we should probably GPG-sign it or something. Furthermore, this image can be checked in to the repo. Launching a VM-based keep would involve pasting the image into the address space.

haraldh:
>  Launching a VM-based keep would involve pasting the image into the address space.

~~pasting~~ `mmap`

But yeah... we reached the point, where the final product could just mmap an image for the first mem slot. :partying_face:

connorkuehl (approved):

src/main.rs
```diff
@@ -115,11 +115,19 @@ struct Exec {
     code: PathBuf,
 }
 
+/// Get a report from a keep
+#[derive(StructOpt)]
+struct Report {
+    /// The payload to run inside the keep
+    code: PathBuf,
+}
```
lkatalin, line 123: For SGX we've got a `Report` type [here](https://github.com/enarx/sgx/blob/master/src/attestation_types/report.rs#L59) already that conforms to the SGX spec. It seems better to use this directly, unless there is some reason we can't.

lkatalin (commented):
See inline.

src/main.rs
connorkuehl: This is a command-line arg struct. `StructOpt` expands this such that one can do `enarx-keepldr report --code "potato.wasm"`.

lkatalin: Ahhh, makes sense. Very good.

lkatalin (approved):

PR #204, opened 2020-12-02
feat(syscall): Add some network syscalls

haraldh:
Implemented proxying of the following syscalls:
    
* socket
* bind
* connect
* accept
* accept4
* listen
* recvfrom

### DIFF
--- a/internal/syscall/src/lib.rs
+++ b/internal/syscall/src/lib.rs
@@ -181,6 +181,10 @@ pub trait SyscallHandler: AddressValidator + Sized {
             ),
             libc::SYS_bind => self.bind(usize::from(a) as _, b.into(), c.into()),
             libc::SYS_listen => self.listen(usize::from(a) as _, usize::from(b) as _),
+            libc::SYS_accept => self.accept(usize::from(a) as _, b.into(), c.into()),
+            libc::SYS_accept4 => {
+                self.accept4(usize::from(a) as _, b.into(), c.into(), usize::from(d) as _)
+            }
 
             SYS_ENARX_GETATT => self.get_attestation(a.into(), b.into(), c.into(), d.into()),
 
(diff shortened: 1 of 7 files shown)


tests/bin/unix_echo.rs
```diff
@@ -0,0 +1,14 @@
+use std::io::{self, Read, Write};
+use std::os::unix::net::{UnixListener, UnixStream};
+
+fn main() -> io::Result<()> {
+    let listener = UnixListener::bind("/tmp/enarx_unix_echo_test_to_test")?;
```
lkatalin, line 5: I might call this something like `/tmp/enarx_unix_echo_to_bin` and `/tmp/enarx_unix_echo_from_bin` (or `echo_from_code` or whatever sounds good) to be a bit more descriptive.

tests/integration_tests.rs
```diff
@@ -226,6 +229,50 @@ fn echo() {
     run_test("echo", 0, input.as_slice(), input.as_slice(), None);
 }
 
+#[test]
+#[serial]
+fn unix_echo() {
+    const FILENAME_IN: &'static str = "/tmp/enarx_unix_echo_test_to_test";
+    const FILENAME_OUT: &'static str = "/tmp/enarx_unix_echo_test_from_test";
```
lkatalin, line 236: Same comment here as `test_to_test` is more ambiguous.

lkatalin (commented):
LGTM, small comment on naming in the test but not a big deal either way.

internal/syscall/src/lib.rs
```diff
@@ -658,4 +678,207 @@ pub trait SyscallHandler: AddressValidator + Sized {
...
+        addr: UntrustedRefMut<u8>,
+        addrlen: UntrustedRefMut<libc::socklen_t>,
+        flags: libc::c_int,
+    ) -> Result {
+        self.trace("accept4", 4);
+
+        if addr.as_ptr().is_null() {
+            return unsafe {
+                self.proxy(
+                    request!(libc::SYS_accept4 => fd, addr.as_ptr(), addrlen.as_ptr(), flags),
```
connorkuehl, line 750: The man page suggests that if `addr` is passed in as null, then a null pointer should be passed in for `addrlen`.

connorkuehl (changes requested):
Please split each syscall into a separate commit

internal/syscall/src/lib.rs
haraldh: @connorkuehl  And? I just pass on what userspace gave me and userspace will get the errors it deserves.

tests/bin/unix_echo.rs
haraldh: thanks! good idea

haraldh:
> Please split each syscall into a separate commit

@connorkuehl as you wish, please re-review
CC @lkatalin

lkatalin (approved):

tests/integration_tests.rs
haraldh: should be resolved

tests/bin/bind.c
```diff
@@ -0,0 +1,29 @@
...
+    };
+    socklen_t sa_len = strlen(UNIX_ABSTRACT_PATH);
+
+    int fd = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
+
+    if (fd < 0)
+        return 1;
+
+    strcpy(sa.sun_path, UNIX_ABSTRACT_PATH);
+    sa.sun_path[0] = '\0';
```
connorkuehl, line 22: Won't this result in the empty string being passed as the path?

tests/bin/bind.c
```diff
@@ -0,0 +1,29 @@
...
+
+    if (bind(fd, (struct sockaddr *)&sa, offsetof(struct sockaddr_un, sun_path) + sa_len) < 0) {
```
connorkuehl, line 24: Should the length field be only `sa_len`?

tests/bin/listen.c
```diff
@@ -0,0 +1,32 @@
...
+
+    if (fd < 0)
+        return 1;
+
+    strcpy(sa.sun_path, UNIX_ABSTRACT_PATH);
+    sa.sun_path[0] = '\0';
+
+    if (bind(fd, (struct sockaddr *)&sa, offsetof(struct sockaddr_un, sun_path) + sa_len) < 0) {
+        return 2;
+    }
```
connorkuehl, line 26: Same questions as `bind` commit.

connorkuehl (approved):
A couple of questions inline, but otherwise LGTM.

tests/bin/bind.c
haraldh: This is an abstract unix socket without a file in the file system.

tests/bin/bind.c
haraldh: sa_len is only the path length

tests/bin/listen.c
haraldh: same

PR #208, opened 2020-12-03
feat(syscall): Report an unknown syscall

haraldh:
Add a method to the `SyscallHandler` to report an unknown syscall,
similar to `trace()`.

### DIFF
--- a/internal/shim-sgx/src/handler.rs
+++ b/internal/shim-sgx/src/handler.rs
@@ -3,6 +3,7 @@
 use crate::Layout;
 
 use core::fmt::Write;
+use primordial::Register;
 use sallyport::{request, Block, Cursor, Request};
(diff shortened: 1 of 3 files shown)


internal/shim-sgx/src/handler.rs
```diff
@@ -145,6 +146,23 @@ impl<'a> SyscallHandler for Handler<'a> {
         self.exit(1)
     }
 
+    #[inline]
```
lkatalin, line 149: Why inline? Just curious.

lkatalin (approved):

internal/shim-sgx/src/handler.rs
haraldh: just a hint that the compiler is allowed to inline

connorkuehl (approved):

PR #213, opened 2020-12-03
feat(shim-sgx): remove build.rs

haraldh:
Use `#[naked]` and `asm!` to recreate what we had with pure assembly
code.

Pro: symbols, constant and `XSave`
Con: no assembler macros

Related: https://github.com/enarx/enarx/issues/819


internal/shim-sgx/src/start.rs
```diff
@@ -0,0 +1,295 @@
...
+    use syscall::SYS_ENARX_ERESUME;
+    use xsave::XSave;
+
+    const GPR: u64 = 4096 - 184;
+    const RSPO: u64 = GPR + 32;
+    const MISC: u64 = GPR - 16;
+    const SRSP: u64 = MISC - 8;
+
+    const STACK: u64 = 9 * 8;
+    const SHIM: u64 = 10 * 8;
```
haraldh, line 18: @lkatalin @npmccallum maybe you can document these and make suggestions how to create this computationally.

internal/shim-sgx/src/enclave.rs
```diff
@@ -0,0 +1,287 @@
...
+    use syscall::SYS_ENARX_ERESUME;
+    use xsave::XSave;
+
+    const GPR: u64 = 4096 - 184;
+    const RSPO: u64 = GPR + 32;
+    const MISC: u64 = GPR - 16;
+    const SRSP: u64 = MISC - 8;
+
+    const STACK: u64 = 9 * 8;
+    const SHIM: u64 = 10 * 8;
```
haraldh, line 27: @npmccallum @lkatalin : maybe you can document these and make suggestions how to create this computationally.

lkatalin: Good suggestion - I wasn't able to dig it up from the original PR and can't remember, but I made a debt issue for this: https://github.com/enarx/enarx-keepldr/issues/218

lkatalin (approved):
This looks right to me, as I compared it with our original. 

Is there a reason not to factor out the repetitive save / load / zero instructions into their own naked asm Rust functions and then call them in `_start()`?

haraldh:
> This looks right to me, as I compared it with our original.
> 
> Is there a reason not to factor out the repetitive save / load / zero instructions into their own naked asm Rust functions and then call them in `_start()`?

hmm... this would involve pushing a return address on the stack... and pop that.. hmm

lkatalin:
> > This looks right to me, as I compared it with our original.
> > Is there a reason not to factor out the repetitive save / load / zero instructions into their own naked asm Rust functions and then call them in `_start()`?
> 
> hmm... this would involve pushing a return address on the stack... and pop that.. hmm

@haraldh Interesting - okay, that does not seem necessary unless you want to do it that way.

connorkuehl (approved):

PR #238, opened 2020-12-09
Refactor VM builder modules

connorkuehl:
The loader is having some growing pains. Let's rearrange some stuff to make it easier to add other stuff.

* Update builder address translation hook to return a closure
* Express syscall blocks as Span for Vm
* Factor out VM construction into modular components
* Move measurement functionality out of Builder::Hook
* Propagate measurement JSON to subcommand so it can print it


haraldh (approved):
nice

npmccallum (approved):

PR #243, opened 2020-12-10
feat(shim-sev): Use multiple sallyport blocks in parallel

haraldh:
This patch enables us to communicate with the Host via multiple
sallyport::Blocks in parallel.

A global HostCallAllocator is added, where one can allocate one
sallyport::Block encapsulated in a `HostCall` type.

This patch also adds the `spin` module, which wraps `spinning::Mutex`
and `spinning::RWLock`, so we can use `impl` on our own types.

### DIFF
--- a/internal/shim-sev/src/allocator.rs
+++ b/internal/shim-sev/src/allocator.rs
@@ -2,7 +2,7 @@
 
 //! The global FrameAllocator
 use crate::addr::{HostVirtAddr, ShimPhysAddr, ShimPhysUnencryptedAddr, ShimVirtAddr};
-use crate::hostcall::HOST_CALL;
+use crate::hostcall::HOST_CALL_ALLOC;
 use crate::{get_cbit_mask, BOOT_INFO};
 
 use nbytes::bytes;
@@ -121,7 +121,7 @@ impl EnarxAllocator {
         };
 
         let meminfo = {
-            let mut host_call = HOST_CALL.try_lock().unwrap();
+            let mut host_call = HOST_CALL_ALLOC.try_alloc().unwrap();
             host_call.mem_info().unwrap()
         };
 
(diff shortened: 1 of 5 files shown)


internal/shim-sev/src/allocator.rs
```diff
@@ -242,7 +242,9 @@ impl EnarxAllocator {
                         let new_size: usize = 2u64.checked_mul(last_size as u64).unwrap() as _;
                         let new_size = new_size.min(self.max_alloc);
                         let num_pages = new_size.checked_div(Page4KiB::size() as _).unwrap();
-                        if let Ok(virt_offset) = HOST_CALL.lock().balloon(num_pages) {
+                        if let Ok(virt_offset) =
+                            HOST_CALL_ALLOC.try_alloc().unwrap().balloon(num_pages)
```
lkatalin, line 246: Are we worried about a panic here if the result is `None`?

lkatalin (commented):
I've so far been able to wrap my brain around about half of this and that half looks good; I'll try again soon to understand the rest.

It looks like now instead of just a `HostCall` (which apparently wraps a sallyport block), there's instead a whole array of these called `HostCallAllocator` and there is some mechanism provided to dish them out when they are needed. The `RWLocked` is some locking mechanism on it (seems sensible).

connorkuehl (approved):

internal/shim-sev/src/allocator.rs
connorkuehl: I thought about this during my reading of it too. If all N blocks are busy while syscall N+1 tries to find a block, each syscall will panic when it unwraps None. However, this situation is extremely unlikely. It would require a keep with a ridiculous number of threads. In cases like this during early development, a panic may be more useful and alert us to syscall congestion issues better than a retry mechanism.

Maybe some day when the shim can sleep a process context thread, it can wake it when a block is available.

lkatalin: Thanks @connorkuehl , that makes sense. If we do want it to panic currently, I'd suggest we should get a specific panic output message for debugging if we can. Possibly a debt issue also to change this later on to something like sleep / retry, at least to record that idea.

haraldh: Ack: opened https://github.com/enarx/enarx-keepldr/issues/253

lkatalin (approved):

PR #247, opened 2020-12-11
build(deps): Manual cargo update

haraldh:
Manually ran `cargo update`


connorkuehl:
> Is this something that can or should be automated?

I think it already is. I think Dependabot rate-limits itself on repositories, so it might only open PRs for up to 5 dependencies per day. I think Harald is just expediting this update.

connorkuehl (approved):

PR #254, opened 2020-12-14
Auto-clean Unix sockets created during testing

connorkuehl:
This isn't so much a problem for running tests inside our container, but when developing locally, the unix socket test will fail if the socket is leftover in the /tmp directory:

```console
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }', tests/integration_tests.rs:254:61
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'unix_echo' panicked at 'process `unix_echo` timed out', tests/integration_tests.rs:85:28
```

Consequently, @haraldh would you mind removing this file that belongs to your user: `/tmp/enarx_unix_echo_from_bin` on Rome :-)

### DIFF
--- a/tests/bin/unix_echo.rs
+++ b/tests/bin/unix_echo.rs
@@ -1,14 +1,55 @@
 use std::io::{self, Read, Write};
 use std::os::unix::net::{UnixListener, UnixStream};
 
+use std::path::{Path, PathBuf};
+
+struct TidySocket<T> {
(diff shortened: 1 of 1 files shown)


haraldh:
I would prefer https://github.com/enarx/enarx-keepldr/pull/256

Issue #260, opened 2020-12-17
Add integration tests for syscalls


haraldh:
The following syscalls need integration tests:
* [ ] pipe
* [ ] fcntl with F_GETFD, F_SETFD, F_GETFL, F_SETFL
* [ ] ioctl with FIONBIO
* [ ] getsockname
* [ ] sendto
* [ ] setsockopt
* [ ] epoll_create1
* [ ] epoll_ctl
* [ ] epoll_wait
* [ ] getpid

lkatalin:
@haraldh I've folded these into our master list of syscalls that need testing: https://github.com/enarx/enarx-keepldr/issues/9. Is it okay to close this issue?

PR #261, opened 2020-12-17
Add various syscalls for networking

haraldh:
* pipe
* fcntl with F_GETFD, F_SETFD, F_GETFL, F_SETFL
* ioctl with FIONBIO
* getsockname
* sendto
* setsockopt
* getpid

The dept issue for the integration_tests is: https://github.com/enarx/enarx-keepldr/issues/260

### DIFF
--- a/internal/syscall/src/lib.rs
+++ b/internal/syscall/src/lib.rs
@@ -204,6 +204,13 @@ pub trait SyscallHandler: AddressValidator + Sized {
                 e.into(),
                 f.into(),
             ),
+            libc::SYS_setsockopt => self.setsockopt(
+                usize::from(a) as _,
+                usize::from(b) as _,
+                usize::from(c) as _,
+                d.into(),
+                usize::from(e) as _,
+            ),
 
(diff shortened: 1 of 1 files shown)


npmccallum (changes requested):
Overall, LGTM.

However, the implementation of `epoll_ctl()` is naive and has a security hole. `epoll_data_t` **MUST NOT** be leaked to the host. A naive attempt to fix this would be to store it internally and replace it with an opaque index to the host. But this would still allow a malicious host to select a different pointer for use inside the keep.

Instead, we need to create a structure to remember the specific events the data is registered for and zero the data field when we send it to the host. That is, we need to mimic the Linux behavior here.

npmccallum:
@haraldh I suggest moving epoll stuff to a separate PR and I can merge the rest.

haraldh:
> @haraldh I suggest moving epoll stuff to a separate PR and I can merge the rest.

@npmccallum done

npmccallum (approved):

connorkuehl (approved):

PR #264, opened 2020-12-18
refactor(syscall): split up the SyscallHandler trait

haraldh:
The syscall crate lib.rs grows bigger and bigger.

This patch splits up the syscalls and functions in sub-traits.
It does not change any functions and implementations.


lkatalin (approved):
Overall, I like this. Thanks for doing it!

Is there a reason that `BaseSyscallHandler` and `SyscallHander` are two different things? They seem similar to me.

I also wonder if there would be a better way to organize the giant list of syscalls in `SyscallHandler::syscall()` based on how they are split up among the different handlers.

haraldh:
I factored out BaseSyscall because I was not splitting the gigantic match function. 
Splitting that function would still include the pattern... but yeah... it's an idea, which might half it in the end.

haraldh:
Ah... I forgot... BaseSyscall is needed by all the other traits, because of e.g. `trace` and `cursor`.

And SyscallHandler collects all the components.

haraldh:
> I also wonder if there would be a better way to organize the giant list of syscalls in `SyscallHandler::syscall()` based on how they are split up among the different handlers.

Well, there has to be one primary `match`... but you are right... we could also call to subcomponents for the final call with the casts.

lkatalin:
> > I also wonder if there would be a better way to organize the giant list of syscalls in `SyscallHandler::syscall()` based on how they are split up among the different handlers.
> 
> Well, there has to be one primary `match`... but you are right... we could also call to subcomponents for the final call with the casts.

I was just thinking to reorder the lines and add comments maybe, not change any logic. But I also think it's okay the way it is.

haraldh:
> > > I also wonder if there would be a better way to organize the giant list of syscalls in `SyscallHandler::syscall()` based on how they are split up among the different handlers.
> > 
> > 
> > Well, there has to be one primary `match`... but you are right... we could also call to subcomponents for the final call with the casts.
> 
> I was just thinking to reorder the lines and add comments maybe, not change any logic. But I also think it's okay the way it is.

@lkatalin reordered the lines

lkatalin (approved):

npmccallum (approved):

Issue #205, opened 2020-12-02
shim-sev: Implement munmap and deallocate pages


haraldh:
The allocator needs deallocate/free for `munmap` to be useful. Better free the memory, if we don't want to run out of it.

PR #265, opened 2020-12-18
shim-sev: Introduce a new allocator and the munmap syscall

haraldh:
*   feat(shim-sev): switch to a linked list allocator
    
    The old allocator was a bump allocator, which could not free unused memory.
    
    This patch uses the linked_list_allocator crate to back the management
    of the free memory.
    
    The mapping of the host virtual addresses to shim physical addresses was moved
    into its own `HostMap` type.
    
*   feat(shim-sev): implement the munmap syscall
    
    The old unmap syscall was only a stub, because the old allocator
    could not free memory.
    
    Resolves: https://github.com/enarx/enarx-keepldr/issues/205

### DIFF
--- a/internal/shim-sev/src/hostmap.rs
+++ b/internal/shim-sev/src/hostmap.rs
@@ -1,0 +1,231 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Provides a mapping of the shim physical addresses to the hypervisor host virtual addresses,
+//! needed for proxying syscalls.
+//!
+//! With the mapping the shim can construct a syscall for the kernel with the address
+//! space of the host/hypervisor. Otherwise the hypervisor would have to translate
+//! the shim physical addresses and would have to parse and understand
+//! every syscall and correct the addresses.
+//!
+//! Because of memory ballooning (requesting more memory on demand) the hosts memory
+//! is not contiguous and therefore we need a map for every memory region.
+//!
+//! ```
+//!                   Host             Shim
+//!                  virtual          physical
+//!    0x7fee0000  +--------+------>+--0x0000--+
+//!                |        |       |          |
+//!                |        |       |          |
+//!                |        |       |          |
+//!    0x7fee0400  +--------+---+-->+  0x0400  |        Payload
+//!                             |   |          |        virtual
+//!    0x3abb0000  +--------+---+   |          +<-----+---------+ 0x44450000
+//!                |        |       |          |      |         |
+//!                |        |       |          |      |         |
+//!                |        |       |          |      |         |
+//!                |        |   +-->+  0x0a00  |      |         |
+//!                |        |   | | |          |      |         |
+//!    0x3abb0600  +--------+---+ | |          |      |         |
+//!                               | |          |      |         |
+//!    0x5ccd0000  +--------+-----+ |          |      |         |
+//!                |        |       |          |      |         |
+//!                |        |       |          +<-----+---------+ 0x44450a00
+//!                |        |       |          |
+//!                |        |    +--+  0x1300  |
+//!                |        |    |  |          |
+//!                |        |    |  |          |
+//!    0x5ccd0700  +--------+----+  |          |
+//!                                 |          |
+//!                                 |          |
+//! ```
+use crate::addr::{HostVirtAddr, ShimPhysUnencryptedAddr};
+
+use crate::print::PrintBarrier;
+use crate::spin::RWLocked;
+use core::alloc::Layout;
+use core::mem::{align_of, size_of};
+use linked_list_allocator::Heap;
+use lset::{Line, Span};
+use primordial::{Address, Page as Page4KiB};
+use spinning::Lazy;
+
+/// The global [`HostMap`](HostMap) RwLock
+pub static HOSTMAP: Lazy<RWLocked<HostMap>> =
+    Lazy::new(|| RWLocked::<HostMap>::new(HostMap::new()));
+
+struct HostMemListPageHeader {
+    next: Option<&'static mut HostMemListPage>,
+}
+
+/// A physical memory region with the corresponding host virtual start address
+#[derive(Clone, Copy)]
+struct HostMemEntry {
+    shim: Line<usize>,
+    virt_start: usize,
+}
+
+/// Number of memory list entries per page
+pub const HOST_MEM_LIST_NUM_ENTRIES: usize = (Page4KiB::size()
+    - core::mem::size_of::<HostMemListPageHeader>())
+    / core::mem::size_of::<HostMemEntry>();
+
+struct HostMemListPage {
+    header: HostMemListPageHeader,
+    ent: [HostMemEntry; HOST_MEM_LIST_NUM_ENTRIES],
+}
+
+/// A mapping of the shim physical addresses to the hypervisor host virtual addresses
+pub struct HostMap {
+    num_pages: usize,
+    host_mem: HostMemListPage,
+}
+
+impl HostMap {
+    fn new() -> Self {
+        HostMap {
+            num_pages: 0,
+            host_mem: HostMemListPage {
+                header: HostMemListPageHeader { next: None },
+                ent: [HostMemEntry {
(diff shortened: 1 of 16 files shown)


internal/shim-sev/src/hostmap.rs
```diff
@@ -0,0 +1,204 @@
...
+                }; HOST_MEM_LIST_NUM_ENTRIES],
+            },
+        }
+    }
+
+    fn get_virt_offset(&self, addr: usize) -> Option<i64> {
+        let mut free = &self.host_mem;
+        loop {
+            for i in free.ent.iter() {
+                if i.start == 0 {
```
lkatalin, line 70: Trying to wrap my head around how all of this works together as I haven't looked at the virtual address components at all. When does the `HostMemEntry.start` get updated to hold a non-zero value after `HostMap::new()`? Is there an example I can see of how this part will be used? I get why it should be a linked list and how `NUM_ENTRIES` is determined but I think I'm missing something about what the data in `HostMemEntry` means.

internal/shim-sev/src/syscall.rs
```diff
@@ -377,8 +377,20 @@ impl SyscallHandler for Handler {
...
         self.trace("munmap", 2);
+
+        let addr = addr.validate_slice(lenght, self).ok_or(libc::EINVAL)?;
+
+        ALLOCATOR
+            .write()
+            .unmap_memory(
+                SHIM_PAGETABLE.write().deref_mut(),
+                VirtAddr::from_ptr(addr.as_ptr()),
+                lenght,
```
lkatalin, line 390:
```suggestion
                length,
```

internal/shim-sev/src/hostmap.rs
```diff
@@ -0,0 +1,204 @@
...
+/// The global [`HostMap`](HostMap) RwLock
+pub static HOSTMAP: Lazy<RWLocked<HostMap>> =
+    Lazy::new(|| RWLocked::<HostMap>::new(HostMap::new()));
+
+struct HostMemListPageHeader {
+    next: Option<&'static mut HostMemListPage>,
+}
+
+#[derive(Clone, Copy)]
+struct HostMemEntry {
```
lkatalin, line 28: Any comment about what this struct's data means?

internal/shim-sev/src/syscall.rs
haraldh: thanks, corrected

internal/shim-sev/src/syscall.rs
```diff
@@ -377,8 +377,20 @@ impl SyscallHandler for Handler {
         }
     }
 
-    fn munmap(&mut self, _addr: UntrustedRef<u8>, _lenght: usize) -> sallyport::Result {
+    fn munmap(&mut self, addr: UntrustedRef<u8>, lenght: usize) -> sallyport::Result {
         self.trace("munmap", 2);
+
+        let addr = addr.validate_slice(lenght, self).ok_or(libc::EINVAL)?;
```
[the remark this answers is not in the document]
haraldh: thanks, corrected

internal/shim-sev/src/syscall.rs
```diff
@@ -377,8 +377,20 @@ impl SyscallHandler for Handler {
         }
     }
 
-    fn munmap(&mut self, _addr: UntrustedRef<u8>, _lenght: usize) -> sallyport::Result {
+    fn munmap(&mut self, addr: UntrustedRef<u8>, lenght: usize) -> sallyport::Result {
```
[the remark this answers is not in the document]
haraldh: thanks, corrected

internal/shim-sev/src/hostmap.rs
haraldh: Changed the logic to recognize `end == 0` as an uninitialized element, because it is that's how the initial memory contents is set to with `page.write_bytes(0, 1)`.

Added more comments in the hope it clarifies the logic.

internal/shim-sev/src/hostmap.rs
haraldh: Added more comments in the hope it clarifies the logic.

internal/shim-sev/src/hostmap.rs
```diff
@@ -0,0 +1,241 @@
...
+
+                            (*last_page).header.next = Some(&mut *new_page);
+                            this.num_pages = this.num_pages.checked_add(1).unwrap();
+                            new_page
+                        }
+                        Some(ref mut p) => *p as *mut _,
+                    };
+                }
+            }
+        }
```
connorkuehl, line 173: I think it's harder to read a function when there's another function inside it larger than a reasonably simple lambda. I would suggest just making it a private method within the impl block instead, or just put its contents between the unsafe fences.

internal/shim-sev/src/hostmap.rs
```diff
@@ -0,0 +1,241 @@
...
+        // memory in `inner_extend_slots`.
+        unsafe {
+            SHIM_CAN_PRINT.store(false, Ordering::Release);
+        }
+
+        inner_extend_slots(self.write(), mem_slots, allocator);
+
+        unsafe {
+            SHIM_CAN_PRINT.store(true, Ordering::Release);
+        }
```
connorkuehl, line 186: Suggestion: a `PrintBarrier` type which stores `false` when it's created and stores `true` in `SHIM_CAN_PRINT` when it is dropped.

internal/shim-sev/src/hostmap.rs
```diff
@@ -0,0 +1,241 @@
...
+                last_end = i.shim.end;
+            }
+
+            // we have reached the end of the free slot page
+            // advance to the next page
+            if let Some(f) = free.header.next.as_deref_mut() {
+                free = f;
+            } else {
+                return None;
+            }
```
connorkuehl, line 238: This can be a constant-time operation if the hostmap carries a "next available index" field. In order to avoid gaps when removing a mapping it'll be necessary to replace the "to-be-deleted" element with the final element in the list and then update the "next available index" field accordingly.

connorkuehl (changes requested):

internal/shim-sev/src/hostmap.rs
haraldh: Hmm, ballooning doesn't happen that often and I think it is time critical here. Also, I don't think we will support memory shrinking.

Happy to review and accept PRs, though :)

internal/shim-sev/src/hostmap.rs
haraldh: If this pattern will occur more often, then I will come back to this suggestion.

haraldh: Although, introducing it might help others to not make mistakes.

internal/shim-sev/src/hostmap.rs
haraldh: ok

internal/shim-sev/src/hostmap.rs
haraldh: Done.. Thanks!

internal/shim-sev/src/hostmap.rs
haraldh: Done.. Thanks!

haraldh:
@connorkuehl @lkatalin please re-review

connorkuehl (approved):

internal/shim-sev/src/hostmap.rs
```diff
@@ -0,0 +1,231 @@
...
+//!                |        |       |          |      |         |
+//!                |        |       |          +<-----+---------+ 0x44450a00
+//!                |        |       |          |
+//!                |        |    +--+  0x1300  |
+//!                |        |    |  |          |
+//!                |        |    |  |          |
+//!    0x5ccd0700  +--------+----+  |          |
+//!                                 |          |
+//!                                 |          |
+//! ```
```
lkatalin, line 41: :+1: this diagram

lkatalin (approved):

PR #273, opened 2021-01-04
Take clippy suggestion to use &Path instead of &PathBuf

connorkuehl:
Otherwise CI fails with its newly updated clippy definitions


lkatalin (approved):
We must appease clippy.

haraldh (approved):

PR #296, opened 2021-01-22
Update for new SGX API

npmccallum:

### DIFF
--- a/src/backend/sgx/mod.rs
+++ b/src/backend/sgx/mod.rs
@@ -9,7 +9,7 @@ use crate::syscall::{SYS_ENARX_CPUID, SYS_ENARX_ERESUME, SYS_ENARX_GETATT};
 use anyhow::{anyhow, Result};
 use lset::Span;
 use primordial::Page;
-use sgx::enclave::{Builder, Enclave, Entry, Segment};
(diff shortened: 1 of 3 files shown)


connorkuehl (dismissed):

src/backend/sgx/mod.rs
```diff
@@ -186,8 +187,9 @@ impl super::Thread for Thread {
         //
         //   4. Asynchronous exits other than invalid opcode will panic.
         loop {
-            how = match self.thread.enter(how, &mut self.block) {
-                Err(Some(ei)) if ei.trap == Exception::InvalidOpcode => Entry::Enter,
+            registers.rdi = (&mut self.block).into();
```
haraldh, line 190:
```suggestion
            registers.rdx = (&mut self.block).into();
```

haraldh: With this, all works again :+1: good job!

haraldh (changes requested):

src/backend/sgx/mod.rs
npmccallum: Thank you! I was banging my head against a wall on this one...

haraldh (approved):

connorkuehl (approved):

PR #309, opened 2021-01-26
test: add sev_get_att_quote

haraldh:
Inject a fake secret in the unattested launch and test for it in the
integration tests.

### DIFF
--- a/src/backend/sev/unattested_launch.rs
+++ b/src/backend/sev/unattested_launch.rs
@@ -19,9 +19,10 @@
 use std::convert::TryFrom;
 use std::os::unix::net::UnixStream;
 
-use ::sev::launch::Policy;
+use ::sev::launch::{HeaderFlags, Policy};
 use ::sev::session::Session;
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
+use ciborium::value::Bytes;
 use ciborium::{de::from_reader, ser::into_writer};
 use koine::attestation::sev::*;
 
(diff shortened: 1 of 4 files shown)


tests/bin/sev_get_att_quote.c
```diff
@@ -0,0 +1,24 @@
...
+
+    ssize_t size = get_att(nonce, sizeof(nonce), buf, sizeof(buf), &technology);
+
+    if (technology != TEE_SEV)
+        return 0;
+
+    if (size < 0)
+        return 1;
+
+    write(STDOUT_FILENO, buf, size);
```
lkatalin, line 21: I should include this for sgx, too.

lkatalin (approved):

src/backend/sev/unattested_launch.rs
```diff
@@ -46,7 +47,24 @@ pub fn launch(sock: UnixStream) -> Result<()> {
...
+        into_writer(&Bytes::from(ct_vec), &mut ct_enc).context("failed to encode secret")?;
+
+        let secret = session
+            .secret(::sev::launch::HeaderFlags::default(), &ct_enc)
+            .context("gen_secret failed")?;
+
+        Message::Secret(Some(secret))
+    } else {
+        Message::Secret(None)
+    };
```
connorkuehl, line 66: This `else` is unreachable. The `assert!` statement above it guarantees the first branch will always be taken.

One suggestion here would be to effectively replace the `assert!(maches!(...));`

```rust
let msr = match msr {
    Message::Measurement(m) => m,
    _ => panic!("expected measurement packet"),
};

let secret_packet = {
    let session = unsafe { session.mock_verify(msr).context("verify failed")? };
    
    // [...]
};
```

connorkuehl (changes requested):

src/backend/sev/unattested_launch.rs
haraldh: Thanks! Now it looks much better!

connorkuehl (approved):

lkatalin (approved):

PR #316, opened 2021-01-27
Fix C warning

lkatalin:
Apparently the C compiler doesn't like the for loop the other way.

### DIFF
--- a/tests/bin/sgx_get_att_quote.c
+++ b/tests/bin/sgx_get_att_quote.c
@@ -10,7 +10,7 @@ int main(void) {
     int* nonce = NULL;
     unsigned char buf[4598];
     size_t technology;
-    int i = 0;
+    int i;
     unsigned char expected[32] = { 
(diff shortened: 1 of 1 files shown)


tests/bin/sgx_get_att_quote.c
```diff
@@ -35,7 +35,7 @@ int main(void) {
         return 1;
 
     /* check beginning of quote matches expected value */
-    for (i; i < 32; i++)
+    for (; i < 32; i++)
```
haraldh, line 38:
```suggestion
    for (i = 0; i < 32; i++)
```

haraldh (changes requested):

tests/bin/sgx_get_att_quote.c
lkatalin: That looks better to me also.

haraldh (approved):

ueno (approved):

PR #333, opened 2021-02-02
Add network isolation

npmccallum:
This currently requires CAP_SYS_ADMIN. We need to figure out how to
handle this in the future.

Signed-off-by: Nathaniel McCallum <npmccallum@redhat.com>

### DIFF
--- a/src/netlink/interface.rs
+++ b/src/netlink/interface.rs
@@ -1,0 +1,240 @@
+use super::{connection::Connection, Address, Error};
+
+use netlink_packet_route::*;
+
+use std::convert::TryFrom;
+use std::io::ErrorKind;
+use std::net::IpAddr;
+use std::os::unix::io::RawFd;
+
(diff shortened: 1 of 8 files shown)


src/netlink/interface.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+
+        Err(ErrorKind::InvalidData)
+    }
+}
+
+impl Interface {
+    //const IPVLAN_MODE_L2: u16 = 0;
+    //const IPVLAN_MODE_L3: u16 = 1;
```
connorkuehl, line 37: Any future use for these?

haraldh:
As I mentioned, this should probably go into something like [bubblewrap](https://github.com/containers/bubblewrap)

npmccallum:
This PR is superseded by https://github.com/npmccallum/ipvlan .

PR #424, opened 2021-07-12
Split off sgx-heap into enarx-heap

npmccallum:
Signed-off-by: Nathaniel McCallum <nathaniel@congru.us>


ambaxter (approved):
lgtm!

npmccallum:
@haraldh Looks like the GHA runner is offline. I think this is due to us running the container using systemd and it being unable to find the dbus session socket.

```
WARN[0000] The cgroupv2 manager is set to systemd but there is no systemd user session available 
WARN[0000] For using systemd, you may need to login using an user session 
WARN[0000] Alternatively, you can enable lingering with: `loginctl enable-linger 1009` (possibly as root) 
WARN[0000] Falling back to --cgroup-manager=cgroupfs
```

haraldh:
> @haraldh Looks like the GHA runner is offline. I think this is due to us running the container using systemd and it being unable to find the dbus session socket.
> 
> ```
> WARN[0000] The cgroupv2 manager is set to systemd but there is no systemd user session available 
> WARN[0000] For using systemd, you may need to login using an user session 
> WARN[0000] Alternatively, you can enable lingering with: `loginctl enable-linger 1009` (possibly as root) 
> WARN[0000] Falling back to --cgroup-manager=cgroupfs
> ```

fixed

PR #434, opened 2021-08-05
Use a new ELF format for the shim

haraldh:
This commit series will decouple the loader and shim logic of memory layout and communicating that layout.

Get rid of `hostlib`, `BootInfo` and loader centric calculation of the
memory layout, with the help of a linker script, ELF program header entries
and global symbols.

This requires a relative big commit, because it cannot be split without
breaking the CI.

The loader will use these new program header types:

```rust
/// The sallyport program header type
pub const PT_ENARX_SALLYPORT: u32 = PT_LOOS + 0x34a0001;

/// The CR3 pagetable program header type
pub const PT_ENARX_PML4: u32 = PT_LOOS + 0x34a0002;

/// The enarx code program header type
pub const PT_ENARX_CODE: u32 = PT_LOOS + 0x34a0003;
```

These give the loader a hint for the range of that specific memory type,
which require special handling or is used to communicate with the shim.

Resulting ELF headers:

```
❯ readelf -Wl target/debug/build/enarx-keepldr-553d77cbdbfb8a40/out/bin/shim-sev

Elf file type is EXEC (Executable file)
Entry point 0x100000000
There are 12 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x00000000fffa9000 0x00000000fffa9000 0x000000 0x055000 RW  0x1000
  LOAD           0x001000 0x00000000ffffe000 0x00000000ffffe000 0x001000 0x001000 RW  0x1000
  LOAD           0x002000 0x00000000fffff000 0x00000000fffff000 0x001000 0x001000 RW  0x1000
  LOAD           0x003000 0x0000000100000000 0x0000000100000000 0x000166 0x000166 R E 0x1000
  LOAD           0x004000 0x0000000100001000 0x0000000100001000 0x03ae68 0x03ae68 RW  0x1000
  LOAD           0x03f000 0x000000010003c000 0x000000010003c000 0x053680 0x053680 RWE 0x1000
  LOAD           0x093000 0x0000000100090000 0x0000000100090000 0x0039cc 0x0039cc R   0x1000
  LOAD           0x097000 0x0000000100094000 0x0000000100094000 0x002240 0x006a6a RW  0x1000
  DYNAMIC        0x004000 0x0000000100001000 0x0000000100001000 0x000110 0x000110 RW  0x8
  LOOS+0x34a0001 0x001000 0x00000000fffa9000 0x00000000fffa9000 0x000000 0x055000     0x1000
  LOOS+0x34a0002 0x002000 0x00000000fffff000 0x00000000fffff000 0x001000 0x001000     0x1000
  LOOS+0x34a0003 0x000000 0x000000010009b000 0x000000010009b000 0x000000 0x400000     0x1000

 Section to Segment mapping:
  Segment Sections...
   00     .sallyport 
   01     .pml3 
   02     .pml4 
   03     .boot_text 
   04     .dynamic .boot_data .data.rel.ro .got 
   05     .init .text .rela.dyn 
   06     .dynsym .rodata .dynstr .gnu.hash 
   07     .data .got.plt .bss 
   08     .dynamic 
   09     .sallyport 
   10     .pml4 
   11     .code 
```


internal/shim-sev/layout.ld
```diff
@@ -0,0 +1,117 @@
...
+    PT_ENARX_PML4       0x634A0002 FLAGS(0x0ff00000);
+    PT_ENARX_CODE       0x634A0003 FLAGS(0x0ff00000);
+}
+
+/* Configure this to a reasonable size
+   Current block size is 69632, which gives a maximum count of 15420
+   for the 3GB - 4GB range.
+*/
+_ENARX_SALLYPORT_BLOCK_COUNT = 5;
+_ENARX_SALLYPORT_BLOCK_SIZE = 69632;
```
npmccallum, line 31: I might be misremembering. But I thought we had agreed on one PHDR sallyport entry for each separate thread.

npmccallum: Also, it would be helpful for review to have the readelf output from the produced binary available. Perhaps even just a comment in the PR?

internal/shim-sev/src/main.rs
```diff
@@ -38,31 +36,38 @@ mod start;
...
+    /// Extern
+    pub static _ENARX_SALLYPORT_END: u8;
+    /// Extern
+    pub static _ENARX_MEM_START: u8;
+    /// Extern
+    pub static _ENARX_SHIM_START: u8;
+    /// Extern
+    pub static _ENARX_CODE_START: u8;
+    /// Extern
+    pub static _ENARX_CODE_END: u8;
```
npmccallum, line 69: Why are these `u8` instead of their actual types?

src/backend/sev/builder.rs
```diff
@@ -29,6 +29,64 @@ impl kvm::Hook for Sev {
...
+        _vm: &mut VmFd,
+        addr_space: &mut [u8],
+        shim: &Component,
+    ) -> Result<()> {
+        // The initial page tables of the shim must be corrected
+        // with the current CPU's C-bit indicating encrypted memory.
+
+        // query the C-bit location via cpuid
+        use core::arch::x86_64::__cpuid_count;
+        let c_bit_loc = unsafe { __cpuid_count(0x8000_001f, 0x0000_0000) }.ebx;
```
npmccallum, line 43: Technically, this is UB.

src/backend/sev/builder.rs
```diff
@@ -29,6 +29,64 @@ impl kvm::Hook for Sev {
...
     }
 
+    fn shim_loaded(
+        &mut self,
+        _vm: &mut VmFd,
+        addr_space: &mut [u8],
+        shim: &Component,
+    ) -> Result<()> {
+        // The initial page tables of the shim must be corrected
+        // with the current CPU's C-bit indicating encrypted memory.
```
npmccallum, line 39: In the case where the tenant has provided the shim over the network, how does the tenant code calculate the resulting measurement?

internal/shim-sev/src/main.rs
haraldh: I don't want rust to have any assumptions about alignment.

src/backend/sev/builder.rs
haraldh: @npmccallum: why?

@connorkuehl: that was your code :)

src/backend/sev/builder.rs
haraldh: Well, good question. So start with 16-bit mode?

internal/shim-sev/layout.ld
haraldh: > I might be misremembering. But I thought we had agreed on one PHDR sallyport entry for each separate thread.

In my model all CPUs/threads can use any free sallyport block. Allocating 512 blocks is waaaay too much and takes minutes on Milan to measure upfront. I would rather allocate additional blocks on demand, if a new CPU is started.

src/backend/sev/builder.rs
npmccallum: @haraldh Because issuing the CPUID instruction without determining whether or not the instruction is supported by the processor results in UB. This is a complicated problem. See my comment [here](https://github.com/enarx/sgx/blob/main/src/types/secs.rs#L66). And [here](https://github.com/rust-lang/rust/issues/60123#issuecomment-581621838).

Now, any CPU which supports SEV will also, by definition, support CPUID. But it is still potentially UB.

internal/shim-sev/layout.ld
haraldh: > Also, it would be helpful for review to have the readelf output from the produced binary available. Perhaps even just a comment in the PR?

Added to the top comment

src/backend/sev/builder.rs
npmccallum: Maybe. Let's solve the problem later.

internal/shim-sev/src/main.rs
npmccallum: Don't you just force these assumptions on later code?

internal/shim-sev/layout.ld
npmccallum: It takes minutes using SEV on Milan. But we won't do that.

However, if we're going to commit to only starting one CPU up front, then we should only allocate one sallyport up front.

connorkuehl:
@haraldh If you split these commits out into their own rolled-up PR, I think they're mergeable and don't have to get hung up on the meat of this PR:

* feat(shim-sev): rebuild shim-sev on linker script change
* feat(shim-sev): add debug function
* fix(shim-sev): `_enarx_asm_triple_fault` correct addresses
* fix: update to sallyport master
* feat(shim-sev): impl `Default` for `EncPhysOffset`
* fix(shim-sev): do an early exit in `stack_trace()`

src/backend/sev/builder.rs
haraldh: We can exchange information about the c-bit in the client<->server communication and the client could self measure then with the same c-bit setup. This function is called before the `measurement`, so if we make c-bit a variable, this should work.

haraldh:
> @haraldh If you split these commits out into their own rolled-up PR, I think they're mergeable and don't have to get hung up on the meat of this PR:
> 
>     * feat(shim-sev): rebuild shim-sev on linker script change
> 
>     * feat(shim-sev): add debug function
> 
>     * fix(shim-sev): `_enarx_asm_triple_fault` correct addresses
> 
>     * fix: update to sallyport master
> 
>     * feat(shim-sev): impl `Default` for `EncPhysOffset`
> 
>     * fix(shim-sev): do an early exit in `stack_trace()`

I don't need them now. I will work on `sallyport` in the meantime.

internal/shim-sev/layout.ld
haraldh: 2... one for debug print while filling out the other.

src/backend/sev/builder.rs
haraldh: So, shall we copy&paste `core::arch::x86_64::has_cpuid()` and use it in our code?

internal/shim-sev/src/main.rs
haraldh: Ok, done.

internal/shim-sev/layout.ld
haraldh: changed

src/backend/sev/builder.rs
connorkuehl: This code is unreachable without having passed all of the SEV checks which would guarantee this cpuid is correct.

npmccallum (approved):

PR #436, opened 2021-08-09
Enable normal SEV on Milan 

haraldh:
These commits lets milan pass our test suite with non-SNP SEV.

### DIFF
--- a/internal/shim-sev/src/syscall.rs
+++ b/internal/shim-sev/src/syscall.rs
@@ -40,7 +40,6 @@ struct X8664DoubleReturn {
 /// # Safety
 ///
 /// This function is not be called from rust.
-#[inline(never)]
 #[naked]
 pub unsafe extern "sysv64" fn _syscall_enter() -> ! {
     use crate::gdt::{USER_CODE_SEGMENT, USER_DATA_SEGMENT};
--- a/src/backend/sev/builder.rs
+++ b/src/backend/sev/builder.rs
@@ -47,6 +47,7 @@ impl kvm::Hook for Sev {
         let chain_packet = match generation {
             sev::Generation::Naples => Message::CertificateChainNaples(chain),
             sev::Generation::Rome => Message::CertificateChainRome(chain),
+            sev::Generation::Milan => Message::CertificateChainMilan(chain),
         };
         into_writer(&chain_packet, &self.0)?;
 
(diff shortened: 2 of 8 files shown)


src/backend/sev/builder.rs
```diff
@@ -47,6 +47,7 @@ impl kvm::Hook for Sev {
         let chain_packet = match generation {
             sev::Generation::Naples => Message::CertificateChainNaples(chain),
             sev::Generation::Rome => Message::CertificateChainRome(chain),
+            sev::Generation::Milan => Message::CertificateChainRome(chain),
```
npmccallum, line 50: Is this supposed to be `CertificateChainMilan`?

internal/shim-sev/src/syscall.rs
```diff
@@ -40,7 +40,7 @@ struct X8664DoubleReturn {
 /// # Safety
 ///
 /// This function is not be called from rust.
-#[inline(never)]
+// #[inline(never)]
```
npmccallum, line 43: Just remove this. It will never be accepted and will turn into a hard error.

npmccallum (changes requested):

internal/shim-sev/src/syscall.rs
haraldh: ok

src/backend/sev/builder.rs
haraldh: It's the same format. Maybe it should be renamed.

npmccallum: It is fine as long as it isn't a mistake.

haraldh: @connorkuehl rename or add `CertificateChainMilan` ?

npmccallum: @haraldh @connorkuehl Can we alias `Milan` => `Rome`? That would make things explicit.

connorkuehl: Hey, sorry, part of this conversation happened in chat. For posterity, here's what I said:

>  to be honest, I was trying to remember why there are different types at all, and I think it boils down to the fact that I needed to force serde to serialize/deserialize a union type by wrapping it in an enum. The important part is that some of the Naples certificates are Small (2048) and the Rome ones are "big" (4096).
> 
> I am partial to adding CertificateChainMilan, even if it is identical to Rome, because as far as the protocol goes, it labels the generation of certificate chain that the backend is sending. We use cbor, but if we used json the overall message type (https://github.com/enarx/koine/blob/master/src/attestation/sev.rs#L32) should come across as something like { "certificate-chain-milan": "blahbinarydata" } which I think is useful information when someone is developing client tooling

Would aliasing result in a Message type like this?: `{ "certificate-chain-rome": "actually a milan binary blob" }`

haraldh: fixed

connorkuehl (approved):

npmccallum (approved):

PR #443, opened 2021-08-13
feat: remove binary::Segment

haraldh:
Do not make another memory copy of the executables
before copying it in the final destination.

Signed-off-by: Harald Hoyer <harald@hoyer.xyz>


src/binary/component.rs
```diff
@@ -96,32 +80,40 @@ impl Component {
...
 
-        Ok(Self {
-            entry: elf.entry as _,
-            segments,
-            pie,
-        })
+        Ok(component)
+    }
+
+    #[allow(dead_code)]
```
npmccallum, line 96: Why?

src/binary/component.rs
```diff
@@ -96,32 +80,40 @@ impl Component {
...
+    pub fn filter_header(&self, type_: u32) -> impl Iterator<Item = &ProgramHeader> {
+        self.elf
+            .program_headers
+            .iter()
+            .filter(move |ph| ph.p_type == type_)
+    }
+
+    #[allow(dead_code)]
```
npmccallum, line 104: Why?

src/binary/component.rs
haraldh: removed

src/binary/component.rs
haraldh: removed

haraldh:
Based on https://github.com/enarx/enarx-keepldr/pull/445

npmccallum (approved):

PR #446, opened 2021-08-16
fix(shim-sev): declare all symbols referenced in asm

haraldh:
In newer rustc versions, those symbols don't reference the global
symbols anymore.

Signed-off-by: Harald Hoyer <harald@hoyer.xyz>


npmccallum:
@haraldh Why does this fail?

haraldh:
> @haraldh Why does this fail?

@npmccallum We need all the fixes for the CI to pass.

PR #460, opened 2021-08-23
Misc cleanups

npmccallum:
Lots of misc cleanups. Both SGX and SEV.


internal/shim-sgx/src/handler.rs
```diff
@@ -66,7 +66,7 @@ impl<'a> Handler<'a> {
             // prevent earlier writes from being moved beyond this point
             core::sync::atomic::compiler_fence(core::sync::atomic::Ordering::Release);
 
-            asm!("syscall");
+            asm!("cpuid");
```
haraldh, line 69: oops.. an oversight for a long time?

src/backend/kvm/vm/mod.rs
```diff
@@ -24,22 +22,31 @@ use kvm_ioctls::{Kvm, VmFd};
...
     kvm: Kvm,
     fd: VmFd,
     regions: Vec<Region>,
     syscall_blocks: Span<VirtAddr, NonZeroUsize>,
-    arch: A,
     _personality: PhantomData<P>,
+    cpus: VecDeque<u64>,
+
+    #[cfg(target_arch = "x86_64")]
+    rip: PhysAddr,
```
haraldh, line 43: This is actually a guest `VirtAddr`.
```suggestion
    rip: VirtAddr,
```

src/backend/sev/personality.rs
```diff
@@ -3,13 +3,16 @@
...
 pub struct Sev;
 
 impl Personality for Sev {
-    fn add_memory(vm: &VmFd, region: &KvmUserspaceMemoryRegion) {
-        mark_encrypted(vm, region).expect("SEV memory pinning failed");
+    fn add_memory(vm: &mut VmFd, region: &KvmUserspaceMemoryRegion) {
+        let enc_region = KvmEncRegion::new(region);
+        ENCRYPT_REGION
+            .ioctl(vm, &enc_region)
+            .map(|_| ())
```
haraldh, line 15: ?

src/backend/sev/runtime.rs
```diff
@@ -1,13 +0,0 @@
...
-use std::os::unix::io::AsRawFd;
-
-use super::ioctl::{KvmEncRegion, KvmUserspaceMemoryRegion, ENCRYPT_REGION};
-
-pub fn mark_encrypted(kvm_fd: &impl AsRawFd, region: &KvmUserspaceMemoryRegion) -> Result<()> {
-    let enc_region = KvmEncRegion::new(region);
-    ENCRYPT_REGION
-        .ioctl(&mut kvm_fd.as_raw_fd(), &enc_region)
-        .map(|_| ())
-}
```
haraldh, line 13: Oops... here it hides... tststs..
@connorkuehl any idea, why this was never part of the sev / sev-iocuddle / snp crate?

haraldh: In the meantime, I reinvented the wheel with https://github.com/enarx/sev-iocuddle/pull/7

haraldh (changes requested):

internal/shim-sgx/src/handler.rs
npmccallum: No. It is a new code. It isn't even a bug. We just need to trigger an exception with an illegal opcode. We can actually use any opcode that is not valid in the case of SGX. But using CPUID here is slightly more self documenting.￼

src/backend/sev/runtime.rs
connorkuehl: It was my hope to keep sev / sev-iocuddle / snp agnostic of any KVM crate. In hindsight, I should have done what you did and passed in a `&[u8]` instead of a `&KvmUserspaceMemoryRegion` and then it would have remained agnostic but the launcher API could have exposed this nicely.

src/backend/kvm/vm/mod.rs
haraldh: ok, I don't care anymore, because it's going to be ripped out soon :)

haraldh (approved):

PR #478, opened 2021-09-07
Rewrite the SGX backend as an elf loader

npmccallum:
Merge these first:
  * https://github.com/enarx/primordial/pull/10
  * https://github.com/rust-osdev/x86_64/pull/303
  * https://github.com/enarx/sallyport/pull/9
  * https://github.com/enarx/sgx/pull/42


internal/shim-sgx/layout.ld
```diff
@@ -0,0 +1,88 @@
+ENTRY(_start)
+
+PHDRS {
+    rodata PT_LOAD FILEHDR PHDRS;
+    data PT_LOAD;
+    text PT_LOAD;
+    dynamic PT_DYNAMIC;
+
+    stk0 PT_LOAD;
+    tcs0 PT_LOAD FLAGS(1 << 20);
```
haraldh, line 10: maybe a comment about `1<<20`

src/binary/component.rs
```diff
@@ -17,10 +17,25 @@ pub enum ComponentType {
...
+/// The enarx init program header type
 pub const PT_ENARX_CODE: u32 = PT_LOOS + 0x34a0003;
 
+/// This segment contains TCS pages.
+#[cfg(feature = "backend-sgx")]
+pub const PF_ENARX_SGX_TCS: u32 = 1 << 20;
+
+/// This segment contains unmeasured pages.
+#[cfg(feature = "backend-sgx")]
+pub const PF_ENARX_SGX_UNMEASURED: u32 = 1 << 21;
```
haraldh, line 29: Where is this used in the linker script?

src/backend/sgx/mod.rs
```diff
@@ -109,91 +126,64 @@ impl crate::backend::Backend for Backend {
         code: Component,
         _sock: Option<&Path>,
     ) -> Result<Arc<dyn Keep>> {
-        // Calculate the memory layout for the enclave.
-        let layout = crate::backend::sgx::shim::Layout::calculate(
-            shim.region().into(),
-            code.region().into(),
-        );
+        // Find the offset for loading the code.
+        let slot = Span::from(shim.find_header(PT_ENARX_CODE).unwrap().vm_range());
```
haraldh, line 130: Maybe add an anyhow context to the error and return with `?` instead of `unwrap()`. This makes debugging a little bit easier.

Same for the rest.

src/binary/component.rs
```diff
@@ -17,10 +17,25 @@ pub enum ComponentType {
 #[cfg(any(feature = "backend-kvm"))]
 pub const PT_ENARX_SALLYPORT: u32 = PT_LOOS + 0x34a0001;
 
-/// The enarx code program header type
-#[cfg(any(feature = "backend-kvm"))]
+/// The enarx init program header type
```
haraldh, line 20: "init"? .. ah, you mean PID 1 .. the only unikernel process. "init" is a little bit ambiguous.

src/binary/component.rs
```diff
@@ -125,4 +140,29 @@ impl<'a> Component<'a> {
...
+            None => return Ok(None),
+        };
+
+        for note in headers {
+            let note = note?;
+            if note.name == name && note.n_type == kind && note.desc.len() == size_of::<T>() {
+                let mut value = MaybeUninit::uninit();
+                let ptr = value.as_mut_ptr() as *mut u8;
+                let slice = from_raw_parts_mut(ptr, size_of::<T>());
+                slice.copy_from_slice(note.desc);
```
haraldh, line 161: Hmm... Is `note.desc` aligned or do we need `value.as_mut_ptr().read_unaligned(note.desc.as_ptr())` ?

src/binary/component.rs
npmccallum: It isn't (yet?). But I want it well defined in case we choose to use this strategy in the shim in the future. That way it won't require a new server side. Only client side changes will be required.

src/backend/sgx/mod.rs
npmccallum: I can do that now. However, my next change is to pull out the common behavior between the backends and make the `Backend` API more intelligent. This was going to include better error handling processes. Do you want to wait for that? Or do you want me to improve error handling now?

src/binary/component.rs
npmccallum: Ugh. Typo.

src/binary/component.rs
npmccallum: I *think* they are equivalent. But I might like your approach better.

internal/shim-sgx/src/entry.rs
```diff
@@ -3,16 +3,15 @@
...
-            options(noreturn)
-        )
+    loop {
+        unsafe {
+            asm!(
+                "syscall",
+                in("rax") libc::SYS_exit,
+                in("rdi") code
+            );
+        }
```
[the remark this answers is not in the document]
npmccallum: Done.

internal/shim-sgx/layout.ld
npmccallum: Fixed.

src/binary/component.rs
npmccallum: Fixed.

src/binary/component.rs
npmccallum: Fixed.

internal/shim-sgx/layout.ld
```diff
@@ -0,0 +1,77 @@
...
+    data PT_LOAD;
+    text PT_LOAD;
+    dynamic PT_DYNAMIC;
+    note PT_NOTE;
+
+    stk0 PT_LOAD;
+    tcs0 PT_LOAD FLAGS(1 << 20); /* PF_ENARX_SGX_TCS */
+    ssa0 PT_LOAD;
+
+    exec 0x634A0003 FLAGS(0); /* PT_ENARX_EXEC */
```
haraldh, line 14: `PT_ENARX_EXEC` or `PT_ENARX_CODE` or `PT_ENARX_INIT` ??

npmccallum: I propose we rename the "code" layer to the executable or "exec" layer.

haraldh (approved):

PR #504, opened 2021-09-27
Rework

npmccallum:
Move the elf loading out of the backends
    
This is a large refactoring. The purpose of this is to standarize how shims are parsed and loaded. This standardized workflow can be used both for hashing the code/exec as well as constructing a keep. I will now describe how this works.
    
The standardized loading procedure is implemented in the `Loader::load()` trait function. This is blanket implemented for all instances of the `Mapper` trait. `Loader::load()` parses the shim/exec and does sanity checking, before passing the parsed version to `Config::new()`. The job of the `Config` instance is to hold configuration information read from the shim binary. Once the `Config` instance is created, it is converted to a `Mapper` instance (i.e. the `Self` of the `Loader` instance which requires `Mapper`). The `Mapper` receives each of the mappings from the elf binary (i.e. the `PT_LOAD` segments). Finally, the `Mapper`, after handling all the mappings, is converted to the output type. Output can be either a `Keep` instance or a hash (i.e. `Vec<u8>`).


wgwoods (approved):
I like the direction here. The ways the traits are designed & used make sense to me and this should make future additions / other uses of the code much easier.

I do think we might need to split or modify the `Backend` trait to make sense in a client/server model, but that can be dealt with later - I filed #506 to track that.

🚢!

PR #505, opened 2021-09-27
Port the KVM backend to the new trait structure

npmccallum:
This commit is incomplete and needs additional work to pass tests. However, it does build.

Note: this PR depends on #504


src/backend/kvm/config.rs
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+
+pub struct Config(());
```
wgwoods, line 5: Could this just be `pub struct Config;`? Or is the `()` necessary because of the `Sized` constraint on `trait Config`?

wgwoods (approved):
I'm gonna let @haraldh review the actual implementation but the general structure/approach looks good to me!

I did have a couple of questions about some code-style quirks but that's more about my own curiosity - those are not dealbreakers.

Get clippy to stop complaining about `note` never being used and I think this looks good to go!

src/backend/kvm/config.rs
wgwoods: ...ah, I've just been informed that this is an idiomatic way of having a zero-sized struct that's `pub`, but can't actually be instantiated outside the crate/module. Kinda threw me off for a minute, and it's not very googleable - could you add a comment here mentioning the reason for the `()`?

npmccallum:
Superseded by #510

PR #511, opened 2021-09-29
Add SEV-SNP support

haraldh:


haraldh:
moved to https://github.com/enarx/enarx/pull/990

PR #512, opened 2021-09-30
Add wasmldr

wgwoods:
This adds `internal/wasmldr` and builds it alongside the shims. With the currently-enabled features, `wasmldr` ends up being ~4.6MB - better than expected, but still room for improvement.

Still TODO:

- [x] figure out how to embed `wasmldr` in the main binary
- [x] figure out how to actually execute it
- [x] figure out how to read in the `.wasm` workload
- [x] hook up stdio
- [x] rig up some tests using `.wasm` modules from `internal/wasmldr/fixtures`
- [x] modify github actions so new integration tests get run


wgwoods:
(note that the current test failures are due to a bug in upstream Rust nightly - see https://github.com/rust-lang/rust/issues/89432)

wgwoods:
Not sure the tests are actually getting run - might need to tweak the github workflows, since they run with "--no-default-features" and don't turn "wasmldr" back on...

wgwoods:
Blerg, the tests inside `internal/wasmldr` are busted now because I moved those test sources to the toplevel `tests/wasm` for use with the wasm integration tests. And I don't know why `build.rs` is failing on the kvm tests. The sgx failures are more interesting - I think that's genuine breakage, not just build-system bugs. Getting close!

wgwoods:
Okay! Fixed up the last couple of things and now everything passes except the SGX tests, which are failing due to something WASI-specific... which probably indicates a bug in the SGX crate... which means these integration tests are already finding new bugs for us to fix!

....hooray?

PR #515, opened 2021-10-04
fix(shim-sev): 128M should be enough for everyone

haraldh:
Increase the maximum allowed executable size for the payload exec.

Signed-off-by: Harald Hoyer <harald@profian.com>


wgwoods (approved):
LGTM - and it makes the wasmldr tests work. 🚢!

